PR #13, opened 2017-10-13
Clean up package dependencies

ddgenome:
Remove dependencies which were never imported.  Add dependencies which
are imported.  Add @types for modules that have them to
devDependencies.  @types/glob-stream seems incorrect so it was not
added.

Tests pass but compilation does not.

@cdupuis @johnsonr @jessitron are you able to assist in eliminating these last few compilations errors?


cdupuis:
@ddgenome thanks for starting this. Moving down the @types like you did breaks every consumer of automation-client-ts as surprisingly not all of them are dev only and the are referenced in the type declarations.

cdupuis:
I removed the API dependencies to some of the more surprising dependencies. This is now good to go.

ddgenome:
I've rebased on master and combined your commits into a single commit. Everything seems to be working.

I'm not sure what you mean by breaking downstream projects. If they are broken, I would guess it's be cause they are not declaring all the dependencies they are using.

Good to merge?

cdupuis:
Yeah, I fixed all those things that broke downstream projects. e.g. logger.d.ts was exporting a winston logger but that was only a dev dependency. So although you weren't using that logger.js in your code, the TS compiler would fail without a winston dependency.

ddgenome:
That surprises me a bit, usually the TS compiler complains if you are using/returning types that are not exported.

cdupuis:
Me too.

PR #18, opened 2017-10-17
Check that your token will work before testing

jessitron:
today i spent too much time trying to run tests with my token before cd told me the one I should use.

This will verify that you have the right token before trying to run the tests, which fail very slowly otherwise.


cdupuis:
Closing this due to inactivity.

PR #21, opened 2017-10-17
The client can't tell us everything

jessitron:


src/internal/transport/websocket/WebSocketClient.ts
```diff
@@ -122,18 +127,30 @@ function register(registrationCallback: () => any, options: WebSocketClientOptio
...
                 } else if (error.response && error.response.status === 400) {
-                    logger.error(`Registration payload for ${registrationPayload.name}` +
-                        `@${registrationPayload.version} was invalid`);
+                    logger.error(`Registration payload for ${nameVersion} was invalid`);
+                    process.exit(1);
+                } else if (error.response
+                    && (error.response.status === 401)) {
+                    const furtherInfo = error.response.data ? `\nFurther information: ${error.response.data}` : "";
+                    logger.error(
+                        `Authorization failed for ${nameVersion} in teams ${registrationPayload.team_ids}` +
```
kipz: Should this be "authentication"?

src/internal/transport/websocket/WebSocketClient.ts
```diff
@@ -122,18 +127,30 @@ function register(registrationCallback: () => any, options: WebSocketClientOptio
...
+                    furtherInfo);
                     process.exit(1);
                 } else if (error.response
-                    && (error.response.status === 401 || error.response.status === 403)) {
-                    logger.error(`Authentication failed for ${registrationPayload.name}` +
-                        `@${registrationPayload.version}`);
+                    && (error.response.status === 403)) {
+                    const furtherInfo = error.response.data ? `\nFurther information: ${error.response.data}` : "";
+                    logger.error(
+                        `Authentication failed for ${nameVersion} in teams ${registrationPayload.team_ids}.
```
kipz: "authorization"?

test/atomist.config.ts
```diff
@@ -45,7 +45,7 @@ class StartUpListener extends AutomationEventListenerSupport {
 export const configuration: Configuration = {
     name: "@atomist/automation-node-tests",
     version: "0.0.6",
-    teamIds: [ "T1L0VDKJP" ],
+    teamIds: "T01234567",
```
ddgenome: Shouldn't this still be an array?

src/internal/transport/websocket/WebSocketClient.ts
```diff
@@ -3,20 +3,25 @@ import axios from "axios";
...
 import { RegistrationConfirmation, WebSocketRequestProcessor } from "./WebSocketRequestProcessor";
 
 export class WebSocketClient {
 
-    constructor(private registrationCallback: () => any,
-                private options: WebSocketClientOptions,
-                private requestProcessor: WebSocketRequestProcessor) {
-
-        register(this.registrationCallback, options, requestProcessor)
+    public static initialize( registrationCallback: () => any,
```
ddgenome: How does this change improve the error message?

src/internal/transport/websocket/WebSocketClient.ts
jessitron: 401 is "HTTP Unauthorized"

test/atomist.config.ts
jessitron: it works either way, but yeah, it might as well be

src/internal/transport/websocket/WebSocketClient.ts
jessitron: it makes the dashboard not open after registration fails.

src/internal/transport/websocket/WebSocketClient.ts
kipz: Yes, it is poorly named:

10.4.2 401 Unauthorized

The request requires user authentication. The response MUST include a WWW-Authenticate header field (section 14.47) containing a challenge applicable to the requested resource. The client MAY repeat the request with a suitable Authorization header field (section 14.8). If the request already included Authorization credentials, then the 401 response indicates that authorization has been refused for those credentials. If the 401 response contains the same challenge as the prior response, and the user agent has already attempted authentication at least once, then the user SHOULD be presented the entity that was given in the response, since that entity might include relevant diagnostic information. HTTP access authentication is explained in "HTTP Authentication: Basic and Digest Access Authentication" [43].

10.4.4 403 Forbidden

The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.

kipz: https://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

kipz: Either way, this is how the backend has been doing it, so this should match.

PR #25, opened 2017-10-18
Add atomist-config and atomist-setup

ddgenome:
Add command to create atomist config.  It will prompt for GitHub
credentials and create a GitHub personal access token and write it to
the config.

Add shell script to clone an Atomist API client repo, call
atomist-config, and start up the client, all in one step.  It can be
called from curl.

Add ability to retain simple console log format on CLIs and in tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,11 @@
     "express-force-ssl": "^0.3.2",
     "express-session": "^1.15.6",
     "fs-extra": "^4.0.2",
+    "github": "^11.0.0",
     "glob-stream": "^6.1.0",
     "graphql": "^0.10.0",
     "graphql-tag": "^2.4.2",
+    "inquirer": "^3.3.0",
     "isomorphic-fetch": "^2.2.1",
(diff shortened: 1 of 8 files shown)


ddgenome:
Pull request auto merge enabled by @null. [atomist:enable-auto-merge]

package.json
```diff
@@ -125,7 +128,8 @@
   "bin": {
     "git-info": "./git.info.js",
     "atomist-cli": "./start.cli.js",
-    "atomist-client": "./start.client.js"
+    "atomist-client": "./start.client.js",
+    "atomist-config": "./atomist-config.js"
```
cdupuis: I guess you had to pick that name with - so that we wouldn't pick it up as configuration, right?

slimslenderslacks:

* 4 successful checks

package.json
ddgenome: I picked it because it made sense to me have the name of the installed script and its source be the same. Is it customary to have script sources named `start.blah.js`?

PR #37, opened 2017-10-21
enhance action chaining; use it for editor chaining

jessitron:


src/operations/generate/UniversalSeed.ts
```diff
@@ -68,5 +68,6 @@ export const RemoveSeedFiles: ProjectEditor = project => {
         ".travis",
     ];
     deleteDirs.forEach(d => defer(project, project.deleteDirectory(d)));
+    // why defer when you're gonna flush after each one
```
johnsonr: Yeah, we could use a different approach now. but this is also fine

src/operations/support/editorUtils.ts
```diff
@@ -22,11 +22,12 @@ import { EditResult, ProjectEditor, successfulEdit } from "../edit/projectEditor
...
-                                               repo: Project,
-                                               editor: ProjectEditor,
-                                               ei: EditMode,
-                                               parameters?: T): Promise<EditResult> {
+    context: HandlerContext,
+    repo: Project,
+    editor: ProjectEditor,
+    ei: EditMode,
+    parameters?: T): Promise<EditResult> {
+    // why bother making a promise?
```
johnsonr: Yes not needed

test/action/actionChainingTest.ts
```diff
@@ -46,6 +47,7 @@ describe("action chaining", () => {
             }).catch(done);
     });
 
+    // this looks identical to the previous test
```
johnsonr: was meant to have different order, maybe it lost that

PR #43, opened 2017-10-24
Persistent directories for project operations

johnsonr:
Uses persistent directories for project operations. This is useful for testing and debugging but also offers potential performance improvement.


src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+import * as os from "os";
+import { CloneDirectoryInfo, CloneOptions, DirectoryManager } from "./DirectoryManager";
+
+export interface StableDirectoryManagerOpts {
+
+    /**
+     * Maximum number of directories to use
+     * @type {number}
+     */
+    maxDirectories?: number;
```
jessitron: if this isn't implemented we should at least put in a comment saying so

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Return undefined if not found
+     */
+    private existingDirectoryFor(user: string, repo: string, branch: string,
+                                 opts: CloneOptions): Promise<CloneDirectoryInfo> {
+        const expectedPath = this.pathFor(user, repo);
+        return fs.pathExists(expectedPath)
+            .then(exists => {
+                if (exists) {
+                    logger.info("%s directories used of %s: Reusing path [%s]",
```
jessitron: wny would we care about that max on existing directories?

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+                logger.info("%s directories used of %s: Returning new path [%s]",
+                    this.directoriesUsed, this.opts.maxDirectories, path);
+                return {
+                    path,
+                    type: "parent-directory" as any,
+                };
+            });
+    }
+
+    private pathFor(user: string, repo: string): string {
```
jessitron: all this "user" is really "owner" right?

jessitron:
@cdupuis Rod wanted you to look at this too. It makes the default checkout behavior be persistent.
This could be enhanced to do fetch instead of clone, which would be faster. It's also handy for testing; you can check out what the automation did after shutting down the client.

@johnsonr right now as this is, what is the behavior when we go to clone a repo, and the directory is already there?

src/project/git/StableDirectoryManager.ts
johnsonr: It's just a useful piece of log info

src/project/git/StableDirectoryManager.ts
johnsonr: Yeah

johnsonr:
@jessitron Right now if the directory is already there it would be reused if the reuse flag is set in options. `GitCommandGitProject` needs to be modified to handle that correctly, so that flag cannot be used yet.

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    private directoriesUsed = 0;
+
+    constructor(pOpts: StableDirectoryManagerOpts) {
+        this.opts = {
+            ...DefaultOpts,
+            ...pOpts,
+        };
+
+        logger.info("Setting up StableDirectoryManager with options %s", JSON.stringify(this.opts));
```
cdupuis: debug? also have you seen `logger.info("Setting up StableDirectoryManager with options %j"`? see https://nodejs.org/api/util.html#util_util_format_format_args. I only found about this recently.

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+
+        if (this.opts.cleanOnExit === true) {
+            logger.info("Registering shutdown hook to delete [%s]", this.opts.baseDir);
```
cdupuis: debug?

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+            registerShutdownHook(() => {
+                logger.info("Cleaning up temporary directories under [%s]", this.opts.baseDir);
```
cdupuis: info ok here.

src/project/git/StableDirectoryManager.ts
cdupuis: debug

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+                }
+            });
+    }
+
+    private freshDirectoryFor(user: string, repo: string, branch: string,
+                              opts: CloneOptions): Promise<CloneDirectoryInfo> {
+        return this.createDir(user, repo)
+            .then(path => {
+                this.directoriesUsed++;
+                logger.info("%s directories used of %s: Returning new path [%s]",
```
cdupuis: debug

src/project/git/StableDirectoryManager.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Default Atomist working directory
+ * @type {string}
+ */
+const AtomistWorkingDirectory = ".atomist-working";
+
+export const PersistentDirectoryManager = new StableDirectoryManager({
+    baseDir: os.homedir() + "/" + AtomistWorkingDirectory,
+    cleanOnExit: false,
```
cdupuis: `${os.homedir()}/${AtomistWorkingDirectory}`

src/project/git/StableDirectoryManager.ts
johnsonr: Thanks!

PR #65, opened 2017-10-31
Test cleanup. Now it passes for me locally

jessitron:
two tests were both creating the same repository, and it deleted it after each of them.
I got 422 on one of them locally, unless I ran it alone.

After a lot of cleanup, it passes locally with all the tests running.

### DIFF
--- a/src/project/git/GitCommandGitProject.ts
+++ b/src/project/git/GitCommandGitProject.ts
@@ -233,11 +233,12 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
 
     public createBranch(name: string): Promise<CommandResult<this>> {
         this.branch = name;
-        return this.runCommandInCurrentWorkingDirectory(`git branch ${name} && git checkout ${name}`);
+        return this.runCommandInCurrentWorkingDirectory(`git branch ${name}`).then(() =>
+            this.runCommandInCurrentWorkingDirectory(`git checkout ${name}`));
     }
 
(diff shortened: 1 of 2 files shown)


src/project/git/GitCommandGitProject.ts
```diff
@@ -227,7 +227,7 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
 
     public createBranch(name: string): Promise<CommandResult<this>> {
         this.branch = name;
-        return this.runCommandInCurrentWorkingDirectory(`git branch ${name}; git checkout ${name}`);
+        return this.runCommandInCurrentWorkingDirectory(`git branch ${name} && git checkout ${name}`);
```
cdupuis: I don't think `&&` works on Windows.

jessitron: good reason

src/project/git/GitCommandGitProject.ts
```diff
@@ -233,11 +233,12 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
 
     public createBranch(name: string): Promise<CommandResult<this>> {
         this.branch = name;
-        return this.runCommandInCurrentWorkingDirectory(`git branch ${name}; git checkout ${name}`);
+        return this.runCommandInCurrentWorkingDirectory(`git branch ${name}`).then(() =>
```
johnsonr: formatting

PR #74, opened 2017-11-07
Better edit tests

jessitron:
my goal here was to test a failed editor, to test the failedEdit convenience function.

which led me to notice that given a custom edit mode, the editor was never called. So I changed that.


johnsonr:
LGTM

PR #78, opened 2017-11-11
Add TypeScript/JavaScript parsing & path expression support

### LINKED ISSUE #77: Add support for parsing and editing JavaScript/TypeScript AST
_Moved from https://github.com/atomist/antlr-ts/issues/4. As TypeScript is core to Atomist, there should be a modest amount of code required, and as we already have a dependency on the TypeScript parser, this support should live here._

Please see the new `ts-parse` branch. This shows the start of adding TypeScript (and JS) support to our `TreeNode` and path expression model. The support should parallel the `MicrogrammarBasedFileParser`, which is why I've copied some of the tests over.

To make this work, we need to get a better grasp on what the TypeScript AST looks like. In particular:

- Extract node names robustly
- Correctly extract positions from TypeScriptElements

johnsonr:
Fixes #77. Also updates `tree-path-ts` version (backward compatible).

Needs more testing, but no reason not to merge at this point as it changes no existing code and adds no dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,9 +19,9 @@
       }
     },
     "@atomist/tree-path": {
-      "version": "0.1.2",
-      "resolved": "https://registry.npmjs.org/@atomist/tree-path/-/tree-path-0.1.2.tgz",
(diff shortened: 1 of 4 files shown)


kristianmandrup:
Wow! Awesome.. I'll have to look into this next week :)

Issue #86, opened 2017-11-13
Add ability to add an empty directory to a Project


johnsonr:
May not be accessible in `InMemoryProject`, but should still be preserved through transformations.

johnsonr:
Still need ability to populate an `InMemoryProject` from a file based project and preserve empty directories

johnsonr:
See pending test

```
it.skip("copies in memory project including empty directory and copies back")
```

ddgenome:
What's the use case here? Git won't respect empty directories.

PR #87, opened 2017-11-13
Project and File enhancements: Starts work on #86

johnsonr:


johnsonr:
Just found a serious bug. Will close until I figure it out.

PR #91, opened 2017-11-13
Allow HandleCommand to return Promise<any>

johnsonr:
Makes `HandleCommand` return `Promise<any | Promise<HandlerResult>`. The union is unnecessary but has value for documentation.

Motivation:

- The status code on `HandleResult` is rarely, if ever, used, vs using `catch`
- Additional data of value on a `HandleResult` instance could just directly come from whatever return type. Note that `handle` methods can carry a type annotation to indicate what they actually promise
- Simple handlers, in particular, are complicated by the need to translate the last promise into a `Promise<HandlerResult>`, when experience has shown that doing that doesn't add much value.

Backward compatible as far as this project is concerned, although technically could break some users (with an easy fix of adding a type annotation to the `handle` method). Most handlers declare they return `Promise<HandleResult`> anyway, and the infrastructure obviously doesn't use the status code, which tells us something.

### DIFF
--- a/test/tree/ast/typescript/typeScriptFileParserProjectTest.ts
+++ b/test/tree/ast/typescript/typeScriptFileParserProjectTest.ts
@@ -25,7 +25,7 @@ describe("TypeScriptFileParser real project parsing: TypeScript", () => {
                     ["TypeScriptFileParser", "TypeScriptAstNodeTreeNode"]);
                 done();
             }).catch(done);
-    });
+    }).timeout(5000);
 
     it("should parse sources from project and use a path expression to find values using convenience method", done => {
         findValues(thisProject, TypeScriptES6FileParser,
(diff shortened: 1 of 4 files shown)


test/tree/ast/typescript/typeScriptFileParserProjectTest.ts
```diff
@@ -46,7 +46,7 @@ describe("TypeScriptFileParser real project parsing: TypeScript", () => {
                 assert(values.length > 2);
                 done();
             }).catch(done);
```
jessitron: tests should `catch(done, done)` so that failures don't have to timeout, and messages are happer

jessitron:
I think this is great! I have an FYI comment about the tests, not a blocker.
This can simplify my samples.

PR #95, opened 2017-11-14
Create a unified atomist CLI

ddgenome:
Move current commands into single CLI.  Organize CLI code into cli
folder.  Disable logging for CLI.  Add help and version command-line
options.  Add compile and install options to start, exiting gracefully
if there is no package.json.


johnsonr:
That was a mistake--Should not have referenced this

PR #96, opened 2017-11-14
fix bug that bothered cloning multiple repos

jessitron:
I used to get errors about "parent directory doesn't exist" when editing multiple projects. This tells it not to delete the whole user directory for each fresh clone, only the repo directory.


johnsonr:
LGTM

Issue #55, opened 2017-10-26
Create unified atomist CLI that has subcommands


ddgenome:
Rather than have multiple scripts that each do one thing, create a unified script that can start the client, login/config, do whatever the CLI does, etc.

johnsonr:
Could we have an interim stage of documenting the present CLI and explaining how people can use that?

ddgenome:
Here's a start: #95 

Still more to do.

ddgenome:
Complete with release of 0.3.4

PR #106, opened 2017-11-19
Run config and git commands in same process

ddgenome:
No not spawn a separate node process for the config and git commands.
Some improvements to yargs usage.  The config command now takes
command-line options and is a lot smarter about handling an existing
config.  Update atomist-setup script to accommodate changes.

Towards #55


src/start.cli.ts
```diff
@@ -109,3 +139,7 @@ function readVersion(): string {
         return "@atomist/automation-client 0.0.0";
     }
 }
+
+/* Local Variables:  */
+/* mode: typescript  */
+/* End:              */
```
cdupuis: What’s that?

ddgenome: Emacs see the node line at the top of the file and wants to use JavaScript mode. This was the easiest way to overrides that. I need to find a better way to address that, so I can remove it if you want.

cdupuis: Yeah, would be. Ice to remove it. Feels like it doesn’t belong there.

Issue #58, opened 2017-10-28
Add caching support to GitCommandGitProject


johnsonr:
Fetch or pull if the repo exists rather than do a fresh clone. If it's dirty, blow it away.

cdupuis:
Also make sure it supports concurrent access.

jessitron:
[ ] If the user does not have the credentials to clone the repo, they should not be able to access it the cached clone.

jessitron:
- [ ] require a fetch to be done within a certain time frame

johnsonr:
There should be an option for in memory caching. We don't need to go to disk for small projects such as most seed repos.

jessitron:
the next step on this is to implement auto-release of locks; that'll complete the write-caching. which may be plenty for this.
For that step, Christian suggested storing the release function on the HandlerContext, which is brilliant because then there's no extra tracking needed, we'll just have the invoker call those functions when they're done. And the release function will need to be wrapped in something that ignores calls after the first one, so it's idempotent. Easy.
The interesting bit will be the editor for the backwards-incompatible change of passing the context into the clone method.

fyi, fetching every time means that the user's creds must have read permission, so that takes care of the piggybacking-security question. also it's easier than deciding when fetch is needed.

cdupuis:
@jessitron what is the state of this? can this be closed?

jessitron:
This is implemented, minimally, but it is not yet the default.

The conversion script for downstream repos is close but doesn't work yet.

If this becomes urgent enough, I can complete it (and update all our clients, plus supply an update editor to other people) in a few days. It hasn't been urgent enough

johnsonr:
This is important. We are probably cloning 6x per push _without modification_ in `github-sdm`.

cdupuis:
Implemented in SDM

PR #108, opened 2017-11-20
Cached repos 58

jessitron:
I am going to create this and then add comments so I don't lose them again.

### DIFF
--- a/src/project/git/GitCommandGitProject.ts
+++ b/src/project/git/GitCommandGitProject.ts
@@ -13,10 +13,15 @@ import { hideString } from "../../internal/util/string";
 import { GitHubRepoRef, isGitHubRepoRef } from "../../operations/common/GitHubRepoRef";
 import { ProjectOperationCredentials } from "../../operations/common/ProjectOperationCredentials";
 import { RemoteRepoRef, RepoRef } from "../../operations/common/RepoId";
-import { CloneOptions, DefaultCloneOptions, DirectoryManager } from "../../spi/clone/DirectoryManager";
+import { CachingDirectoryManager } from "../../spi/clone/CachingDirectoryManager";
+import {
+    CloneDirectoryInfo, CloneOptions, DefaultCloneOptions,
+    DirectoryManager,
+} from "../../spi/clone/DirectoryManager";
 import { StableDirectoryManager } from "../../spi/clone/StableDirectoryManager";
-import { NodeFsLocalProject } from "../local/NodeFsLocalProject";
+import { NodeFsLocalProject, ReleaseFunction } from "../local/NodeFsLocalProject";
 import { GitProject } from "./GitProject";
+import { GitStatus, runStatusIn } from "./gitStatus";
 
 /**
  * Default Atomist working directory
@@ -38,7 +43,7 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
 
     public static fromProject(p: Project, credentials: ProjectOperationCredentials): GitProject {
         if (isLocalProject(p)) {
-            return GitCommandGitProject.fromBaseDir(p.id, p.baseDir, credentials);
+            return GitCommandGitProject.fromBaseDir(p.id, p.baseDir, credentials, () => Promise.resolve());
         }
         throw new Error(`Project ${p.name} doesn't have a local directory`);
     }
@@ -48,11 +53,15 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
      * @param {RepoRef} id
      * @param {string} baseDir
      * @param {ProjectOperationCredentials} credentials
+     * @param release call this when you're done with the project. make its filesystem resources available to others.
+     * @param provenance optional; for debugging, describe how this was constructed
      * @return {GitCommandGitProject}
      */
     public static fromBaseDir(id: RepoRef, baseDir: string,
-                              credentials: ProjectOperationCredentials): GitCommandGitProject {
-        return new GitCommandGitProject(id, baseDir, credentials);
+                              credentials: ProjectOperationCredentials,
+                              release: ReleaseFunction,
+                              provenance?: string): GitCommandGitProject {
+        return new GitCommandGitProject(id, baseDir, credentials, release, provenance);
     }
 
     /**
@@ -66,14 +75,19 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
     public static cloned(credentials: ProjectOperationCredentials,
                          id: RemoteRepoRef,
                          opts: CloneOptions = DefaultCloneOptions,
-                         directoryManager: DirectoryManager = DefaultDirectoryManager): Promise<GitCommandGitProject> {
+                         directoryManager: DirectoryManager = CachingDirectoryManager): Promise<GitCommandGitProject> {
         return clone(credentials, id, opts, directoryManager)
             .then(p => {
-                const pathIntoRepo = !!id.path ?
-                    id.path.startsWith("/") ? id.path : "/" + id.path :
-                    "";
-                const gp = GitCommandGitProject.fromBaseDir(id, p.baseDir + pathIntoRepo, credentials);
-                return gp;
+                if (!!id.path) {
+                    const pathInsideRepo = id.path.startsWith("/") ? id.path : "/" + id.path;
+                    // not sure this will work with cached
+                    const gp = GitCommandGitProject.fromBaseDir(id, p.baseDir + pathInsideRepo, credentials,
+                        () => p.release(),
+                        p.provenance + "\ncopied into one with extra path " + id.path);
+                    return gp;
+                } else {
+                    return p
+                }
             });
     }
 
@@ -83,8 +97,10 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
 
     public newRepo: boolean = false;
 
-    private constructor(id: RepoRef, public baseDir: string, private credentials: ProjectOperationCredentials) {
-        super(id, baseDir);
+    private constructor(id: RepoRef, public baseDir: string,
+                        private credentials: ProjectOperationCredentials, release: ReleaseFunction,
+                        public provenance?: string) {
+        super(id, baseDir, release);
         logger.debug(`Created GitProject with token '${hideString(this.credentials.token)}'`);
     }
 
@@ -102,6 +118,10 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
                     success: commandResult.stdout !== undefined && commandResult.stdout === "",
                 };
             });
+    }
+
+    public gitStatus(): Promise<GitStatus> {
+        return runStatusIn(this.baseDir);
     }
 
     /**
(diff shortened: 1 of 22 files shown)


jessitron:
This is only part of #58. It starts by offering an option to cache cloned directories safely, for write.

The current default:
- re-uses directories
- but clones afresh each time.
- is not safe for simultaneous automation invocations.

This PR gives you the option of a CachedDirectoryManager. It:
- keeps one clone of each repository
- locks the directory on the filesystem so only one automation invocation will use it
- returns temp directories to any automations that request a clone while that one is locked.
- cleans up the clone (fetch, checkout, git clean) before returning it. (actually this code is in GitCommandGitProject.ts, I think it should probably move)

the locking library handles ignoring stale locks from dead processes, but if the automation doesn't release the lock, the lock will stay current until the process exits.

src/project/git/GitCommandGitProject.ts
```diff
@@ -48,11 +53,15 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
      * @param {RepoRef} id
      * @param {string} baseDir
      * @param {ProjectOperationCredentials} credentials
+     * @param release call this when you're done with the project. make its filesystem resources available to others.
```
jessitron: the release function is added so that you can let go of the resource. Most places default it to a no-op

src/project/git/GitCommandGitProject.ts
```diff
@@ -48,11 +53,15 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
+     * @param provenance optional; for debugging, describe how this was constructed
```
jessitron: provenance is for debugging. Optional, but super useful, highly recommend. It helped me figure things out.

src/project/git/GitCommandGitProject.ts
```diff
@@ -69,11 +78,16 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
                          directoryManager: DirectoryManager = DefaultDirectoryManager): Promise<GitCommandGitProject> {
         return clone(credentials, id, opts, directoryManager)
             .then(p => {
-                const pathIntoRepo = !!id.path ?
-                    id.path.startsWith("/") ? id.path : "/" + id.path :
-                    "";
-                const gp = GitCommandGitProject.fromBaseDir(id, p.baseDir + pathIntoRepo, credentials);
-                return gp;
+                if (!!id.path) {
```
jessitron: this is Rod's stuff for supporting projects in subdirs inside Repos. I want to change this a bit but not before this PR is merged.

src/project/git/GitCommandGitProject.ts
```diff
@@ -69,11 +78,16 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
+                    const pathInsideRepo = id.path.startsWith("/") ? id.path : "/" + id.path;
+                    // not sure this will work with cached
+                    const gp = GitCommandGitProject.fromBaseDir(id, p.baseDir + pathInsideRepo, credentials,
+                        () => p.release(),
+                        p.provenance + "\ncopied into one with extra path " + id.path);
+                    return gp;
+                } else {
+                    return p;
```
jessitron: the common case is: clone() now returns a GitCommandGitProject

src/project/git/GitCommandGitProject.ts
```diff
@@ -294,26 +319,73 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
 function clone(credentials: ProjectOperationCredentials,
                id: RemoteRepoRef,
                opts: CloneOptions,
-               directoryManager: DirectoryManager): Promise<GitProject> {
+               directoryManager: DirectoryManager,
+               secondTry: boolean = false): Promise<GitCommandGitProject> {
```
jessitron: clone() retries itself if any of the git commands fail (after getting the directory deleted). This is a recovery mechanism for unexpected repo corruption. It retries only once.

src/project/git/GitCommandGitProject.ts
```diff
@@ -294,26 +319,73 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
     return directoryManager.directoryFor(id.owner, id.repo, id.sha, opts)
         .then(cloneDirectoryInfo => {
-                switch (cloneDirectoryInfo.type) {
-                    case "parent-directory" :
-                        const repoDir = `${cloneDirectoryInfo.path}/${id.repo}`;
-                        const command = (id.sha === "master") ?
-                            `git clone --depth 1 ${id.cloneUrl(credentials)}` :
```
jessitron: some of this moved into a cloneInto() function

src/project/git/GitCommandGitProject.ts
```diff
@@ -294,26 +319,73 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
+                                cloneDirectoryInfo.provenance + "\nRe-using existing clone");
+                        }, error => {
+                            return cloneDirectoryInfo.invalidate().then(() => {
+                                if (secondTry) {
+                                    throw error;
+                                } else {
+                                    return clone(credentials, id, opts, directoryManager, true);
+                                }
                             });
-                    case "actual-directory" :
```
jessitron: this is now supported, but it's called "existing-directory"

src/project/git/GitCommandGitProject.ts
```diff
@@ -294,26 +319,73 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
-                            `git clone ${id.cloneUrl(credentials)}; cd ${id.repo};git checkout ${id.sha}`;
-
-                        logger.info(`Cloning repo '${id.url}' to '${cloneDirectoryInfo.path}'`);
-                        return exec(command, { cwd: cloneDirectoryInfo.path })
-                            .then(_ => {
-                                logger.debug(`Clone succeeded with URL '${id.url}'`);
-                                // fs.chmodSync(repoDir, "0777");
-                                return NodeFsLocalProject.fromExistingDirectory(id, repoDir);
+            switch (cloneDirectoryInfo.type) {
+                case "empty-directory" :
```
jessitron: "parent-directory" is no more; it gives you an "empty-directory"; it set up the directory for you. 
I have forgotten why this was useful but it simplified something a lot.

src/project/git/GitCommandGitProject.ts
```diff
@@ -294,26 +319,73 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
+                case "existing-directory" :
+                    const repoDir = cloneDirectoryInfo.path;
+                    return checkout(repoDir, id.sha)
+                        .then(() => clean(repoDir))
+                        .then(() => {
+                            return GitCommandGitProject.fromBaseDir(id,
+                                repoDir, credentials, cloneDirectoryInfo.release,
+                                cloneDirectoryInfo.provenance + "\nRe-using existing clone");
+                        }, error => {
+                            return cloneDirectoryInfo.invalidate().then(() => {
```
jessitron: here's the part where we retry on unexpected git errors. invalidate() is supplied by the DirectoryManager, and has the effect of deleting the directory (and releasing the lock). goodbye corrupted repo.

src/project/git/GitCommandGitProject.ts
```diff
@@ -294,26 +319,73 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
+                    throw new Error("What is this type: " + cloneDirectoryInfo.type);
+            }
+        });
+}
+
+function cloneInto(credentials: ProjectOperationCredentials,
+                   targetDirectoryInfo: CloneDirectoryInfo, id: RemoteRepoRef) {
+    const repoDir = targetDirectoryInfo.path;
+    const command = (id.sha === "master" && targetDirectoryInfo.transient) ?
+        runIn(".", `git clone --depth 1 ${id.cloneUrl(credentials)} ${repoDir}`) :
```
jessitron: this --depth 1 optimization now runs only for transient directories (temporary ones). If this is our cached clone we want a full one, please.

src/project/git/gitStatus.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface GitStatus {
+    isClean: boolean;
+    raw?: string;
+    sha: string;
+    branch: string;
+    upstream?: {
+        branch: string;
+        inSync: boolean;
+    };
+}
```
jessitron: this is an abstraction to get the git status. It aims to do it using porcelain commands, ones that are not going to change format in future versions of git.
It provides only the information that I needed; it doesn't attempt a full implementation of git status, which trust me is more complicated than you think, and is not necessary.

src/spi/clone/CachingDirectoryManager.ts
```diff
@@ -0,0 +1,110 @@
...
+                        }));
+                }
+            }));
+    },
+};
+
+/*
+ * file locking. only used here
+ */
+
```
jessitron: this could be in a different file but there didn't seem a point. It's just a promise-style wrapper (a deep one) around the locking library.

test/project/git/LockfileLibraryTest.ts
```diff
@@ -0,0 +1,60 @@
+import "mocha";
```
jessitron: exploratory test.

src/project/git/GitCommandGitProject.ts
johnsonr: LGTM

src/project/git/GitCommandGitProject.ts
johnsonr: Nice

src/project/git/GitCommandGitProject.ts
johnsonr: Why not just `GitProject`? This is encouraging callers to couple more deeply. What is the motivation for this change?

src/project/git/GitCommandGitProject.ts
johnsonr: Good

src/spi/clone/StableDirectoryManager.ts
```diff
@@ -52,12 +52,27 @@ export class StableDirectoryManager implements DirectoryManager {
         }
     }
 
+    public invalidate(existing: CloneDirectoryInfo): Promise<void> {
+        return fs.remove(existing.path)
+            .then(() => {
+                logger.info("deleted " + existing.path);
+            }, err => {
+                if (err.code === "ENOENT") {
+                    logger.info("It's already gone");
```
johnsonr: Include the path or delete this message, as it's not useful without it.

src/spi/clone/StableDirectoryManager.ts
```diff
@@ -52,12 +52,27 @@ export class StableDirectoryManager implements DirectoryManager {
...
+                }
+            });
+    }
+
     public directoryFor(owner: string, repo: string, branch: string, opts: CloneOptions): Promise<CloneDirectoryInfo> {
         if (this.opts.reuseDirectories) {
             // Attempt to reuse the directory
             return this.existingDirectoryFor(owner, repo, branch, opts)
-                .then(existing =>
-                    !!existing ? existing : this.freshDirectoryFor(owner, repo, branch, opts));
```
johnsonr: `existing ||` ?

src/spi/clone/StableDirectoryManager.ts
```diff
@@ -109,18 +130,47 @@ export class StableDirectoryManager implements DirectoryManager {
...
+        if (files.length > 0) {
+            return fs.remove(name)
+                .then(() => assureDirectoryExists(name))
+                .catch(err => {
+                    throw new Error("I tried to make this directory be empty: " + name +
+                        " but it didn't work: " + err.message);
+                });
         }
-    }, err => {
-        console.log(`Hope this means 'directory does not exist': ${err}`);
```
johnsonr: I'd make this message more contextual and meaningful or delete it

PR #109, opened 2017-11-20
Fix #34: Add nested parameter support

### LINKED ISSUE #34: Add nested object support to parameters
With new functional style for command handlers, composition rather than inheritance is the best model for parameters. Thus we should be able to handle nested parameters, like this:

```typescript
@Parameters()
class FooParameters {
    @NestedParameter
    barParameters: BarParameters;
}
```

johnsonr:
Makes it easier to reuse parameters via composition rather than getting into inheritance hell.

- Adds nested parameter supports to all command handlers. Handles parameters, mapped parameters and secrets
- Change `generate` utility method to take a `RepoId` explicitly for flexibility and clarity. Backward compatible.
- Introduces new convenience parameter classes for generators

### DIFF
--- a/src/operations/common/params/SourceRepoParameters.ts
+++ b/src/operations/common/params/SourceRepoParameters.ts
@@ -14,7 +14,7 @@ export class SourceRepoParameters {
         description: "owner, i.e., user or organization, of seed repository",
         validInput: GitHubNameRegExp.validInput,
         minLength: 1,
-        maxLength: 50,
+        maxLength: 100,
         required: false,
         displayable: false,
     })
@@ -26,7 +26,7 @@ export class SourceRepoParameters {
         description: "name of the seed repository",
         validInput: GitHubNameRegExp.validInput,
         minLength: 1,
-        maxLength: 50,
+        maxLength: 100,
         required: false,
         displayable: false,
     })
@@ -38,7 +38,7 @@ export class SourceRepoParameters {
         description: "seed repository branch to clone for new project",
         validInput: GitBranchRegExp.validInput,
         minLength: 1,
-        maxLength: 50,
+        maxLength: 256,
         required: false,
         displayable: false,
     })
--- a/src/operations/common/params/SourceRepoParameters.ts
+++ b/src/operations/common/params/SourceRepoParameters.ts
@@ -1,18 +1,12 @@
-import { Parameter } from "../../decorators";
-import { HandlerContext } from "../../index";
-import { Project } from "../../project/Project";
-import { GitHubRepoRef } from "../common/GitHubRepoRef";
-import { GitBranchRegExp, GitHubNameRegExp } from "../common/params/gitHubPatterns";
-import { AbstractGenerator } from "./AbstractGenerator";
+
+import { Parameter } from "../../../decorators";
+import { GitBranchRegExp, GitHubNameRegExp } from "./gitHubPatterns";
 
 /**
- * Support for all seed-driven generators, which start with content
- * in a given repo.
- *
- * Defines common parameters.
- *
+ * Parameters common to anything that works with a single source repo,
+ * such as a seed driven generator
  */
-export abstract class SeedDrivenGenerator extends AbstractGenerator {
+export class SourceRepoParameters {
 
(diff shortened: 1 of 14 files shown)


src/operations/common/params/SourceRepoParameters.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export class SourceRepoParameters {
+
+    @Parameter({
+        pattern: GitHubNameRegExp.pattern,
+        displayName: "Seed Repository Owner",
+        description: "owner, i.e., user or organization, of seed repository",
+        validInput: GitHubNameRegExp.validInput,
+        minLength: 1,
+        maxLength: 50,
```
ddgenome: GitHub's maximum is 100.

src/operations/common/params/SourceRepoParameters.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+    public sourceOwner: string;
+
+    @Parameter({
+        pattern: GitHubNameRegExp.pattern,
+        displayName: "Seed Repository Name",
+        description: "name of the seed repository",
+        validInput: GitHubNameRegExp.validInput,
+        minLength: 1,
+        maxLength: 50,
```
ddgenome: GitHub's maximum is 100.

src/operations/common/params/SourceRepoParameters.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+    public sourceRepo: string;
+
+    @Parameter({
+        pattern: GitBranchRegExp.pattern,
+        displayName: "Seed Branch",
+        description: "seed repository branch to clone for new project",
+        validInput: GitBranchRegExp.validInput,
+        minLength: 1,
+        maxLength: 50,
```
ddgenome: I don't know what the limit is on branch, but I would think we should bump this max to at least 256.

src/operations/common/params/gitHubPatterns.ts
```diff
@@ -0,0 +1,12 @@
+
+export const GitHubNameRegExp = {
```
ddgenome: These look familiar.

src/operations/generate/NewRepoCreationParameters.ts
```diff
@@ -0,0 +1,59 @@
...
+    @MappedParameter(MappedParameters.GitHubOwner)
+    public targetOwner: string;
+
+    @Parameter({
+        pattern: GitHubNameRegExp.pattern,
+        displayName: "Target Repository Name",
+        description: "name of the target repository",
+        validInput: GitHubNameRegExp.validInput,
+        minLength: 1,
+        maxLength: 50,
```
ddgenome: 100

src/operations/generate/NewRepoCreationParameters.ts
```diff
@@ -0,0 +1,59 @@
...
+        required: true,
+        order: 1,
+    })
+    public targetRepo: string;
+
+    @Parameter({
+        displayName: "Project Description",
+        description: "short descriptive text describing the new project",
+        pattern: /.*/,
```
ddgenome: Should this be anchored? We probably no longer require that, right? If we don't require it, there's no reason to do it here.

src/operations/common/params/gitHubPatterns.ts
jessitron: I don't know what to do with this comment?
 are they duplicated somewhere?

johnsonr:
@ddgenome I also have mixed feelings about decorators. However, I think to switch away from them entirely would be too much of a breaking change right now. We could certain add an additional model or provide a convenient way of implementing the lower level metadata interfaces without using decorators.

ddgenome:
Couldn't we just have the decorators populate the `parameters` property instead of the `__parameters` property? It would still require that composition, as opposed to inheritance, but it seems more transparent to me.

johnsonr:
yes, i can't see why we couldn't do that. And that would make life easier. Seems like a separate PR though

ddgenome:
Definitely a separate PR.

PR #111, opened 2017-11-20
Notice the branch passed in

jessitron:
This will fix docker-samples' failure to push. The branch field was only getting defined on init (new repo) or create branch; docker-samples calls push() for an existing branch.

Whether storing branch in this way, separately from the RepoRef, is another question, but at least this makes it work consistently.

### DIFF
--- a/src/project/git/GitCommandGitProject.ts
+++ b/src/project/git/GitCommandGitProject.ts
@@ -102,6 +102,7 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
                         private credentials: ProjectOperationCredentials, release: ReleaseFunction,
                         public provenance?: string) {
         super(id, baseDir, release);
+        this.branch = id.sha;
         logger.debug(`Created GitProject with token '${hideString(this.credentials.token)}'`);
     }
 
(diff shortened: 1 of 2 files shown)


src/project/git/GitCommandGitProject.ts
```diff
@@ -102,6 +102,7 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
                         private credentials: ProjectOperationCredentials, release: ReleaseFunction,
                         public provenance?: string) {
         super(id, baseDir, release);
+        this.branch = id.sha;
```
ddgenome: So the sha is a branch?

jessitron: yeaaaahhh. usually.

what is the name for "sha or branch" ? the git docs seem to use the word "commit" for that, but it's more of a "commit pointer" or "commit identifier" kind of thing

ddgenome: ref?

jessitron: a commit SHA is not technically a ref, i don't think

PR #116, opened 2017-11-21
Fix exec command, use name=value parameters

ddgenome:
Using yargs strict likely broke passing parameters to handlers via the
exec CLI command.  This fixes that, also changing parameter
specification to just NAME=VALUE rather than --NAME=VALUE.

Refactor the exec and start commands to share more code and rely less
on exceptions.  Add some tests.

Towards #55


slimslenderslacks:

* 5 successful checks

PR #117, opened 2017-11-21
Safe cloning

jessitron:
While we're still working on #58 , it's safer to use temporary directories every time than to re-use existing directories without locking.

### DIFF
--- a/src/operations/generate/generatorUtils.ts
+++ b/src/operations/generate/generatorUtils.ts
@@ -7,11 +7,11 @@ import { DefaultDirectoryManager } from "../../project/git/GitCommandGitProject"
 import { LocalProject } from "../../project/local/LocalProject";
 import { NodeFsLocalProject } from "../../project/local/NodeFsLocalProject";
 import { Project } from "../../project/Project";
+import { DirectoryManager } from "../../spi/clone/DirectoryManager";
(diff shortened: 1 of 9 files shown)


jessitron:
ok no this is harder than I thought because the generator utils are coupled with StableDirectoryManager. Gotta revamp that.

src/operations/generate/generatorUtils.ts
```diff
@@ -41,21 +42,21 @@ export function generate(startingPoint: Promise<Project> | Project,
...
-            .then(proj => proj.deleteDirectory(".git"))
-            .then(independentCopy => toEditor<object>(editor)(independentCopy, ctx, params))
-            .then(r => r.target)
-            .then(populated => {
-                logger.debug("Persisting repo at [%s]: owner/repo=%s:%s",
-                    (populated as LocalProject).baseDir, targetId.owner, targetId.repo);
-                return persist(populated, credentials, targetId);
-            }));
+
+    return TmpDirectoryManager.directoryFor(targetId.owner, targetId.repo, "master", {})
```
johnsonr: Couldn't the `DirectoryManager` instance to use be the final argument with a default value. That would avoid the hard coding (which I realize you didn't introduce)

PR #118, opened 2017-11-22
Fix error in attempting to commit with git project isn't dirty

johnsonr:
Motivation was to fix an error in attempting to commit when git project isn't dirty. This turned into an overhaul of `EditResult.edited` which now has 3 possible values, depending on the editor:

- `false`: Editor says I didn't edit this or don't want you to persist any changes I made
- `true`: Editor says I did edit this, don't bother checking with git
- `undefined`: Editor says I have no idea what I did. Likely to result from wrapping a simple function returning `Promise<Project>`

Paired with @jessitron

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


jessitron:
"jim", you were slow on this one, i merged it myself

PR #120, opened 2017-11-23
Remove Java and Spring automations. Now in atomist/spring-automations

johnsonr:
Java and Spring automations do not belong in this repository. They are now in `atomist/spring-automations`.

NB: Downstream projects relying on some of this functionality, such as `atomist-blogs/spring5-kotlin`, should now depend on `spring-automations`. *Therefore this will need to publish to npm.*


test/operations/generate/universalSeedMetadataTest.ts
```diff
@@ -36,18 +34,6 @@ describe("UniversalSeed metadata test", () => {
         validateUniversalSeedMetadata(md);
     });
 
-    it("supports metadata inheritance: one level", () => {
```
ddgenome: Not necessary for this PR, but it may be a good idea to retain an "inherited" generator in the tests so we can retain tests for this functionality.

PR #129, opened 2017-11-28
Release resources

jessitron:
The objective is for the cached directory manager to attach a release step to the context, and have it invoked automatically when the invocation finishes.

### DIFF
--- a/src/internal/invoker/resourceRecovery.ts
+++ b/src/internal/invoker/resourceRecovery.ts
@@ -1,0 +1,35 @@
+import { logger } from "../util/logger";
+
+export interface ReleaseStep { how: () => Promise<void>; what: string; }
+
+export interface ResourceRecovery {
+    releaseSteps?: ReleaseStep[];
+}
+
+export function addReleaseStep(ctx: ResourceRecovery, release: ReleaseStep): void {
+    if (!ctx.releaseSteps) { // this is atomic in Node, right?
+        ctx.releaseSteps = [];
+    }
+    ctx.releaseSteps.push(release);
+}
+
+export function callReleaseSteps(ctx: ResourceRecovery): Promise<void> {
+    if (ctx.releaseSteps) {
+        function both(f1: ReleaseStep, f2: ReleaseStep) {
+            return {
+                how: () => f1.how()
+                    .then(() => f2.how()
+                        .catch(error => {
+                            logger.warn("Failed to release resource %s: %s", f2.what, error);
+                        }))
+                , what: f1.what + " and " + f1.what,
+            };
+        }
+
+        const releaseEverything = ctx.releaseSteps
+            .reduce(both, { how: () => Promise.resolve(), what: "inconceivable" });
+        return releaseEverything.how();
+    } else {
+        return Promise.resolve();
+    }
+}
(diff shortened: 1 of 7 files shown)


src/internal/transport/AbstractRequestProcessor.ts
```diff
@@ -170,7 +179,8 @@ export abstract class AbstractRequestProcessor implements RequestProcessor {
                 .catch(err => {
                     this.handleCommandError(err, command, ci, ctx, callback);
                 });
-        } catch (err) {
+        } catch
+            (err) {
```
ddgenome: Why the newline?

src/internal/metadata/metadataReading.ts
```diff
@@ -22,7 +22,7 @@ export function metadataFromInstance(h: any): CommandHandlerMetadata | EventHand
     if (isEventHandlerMetadata(h)) {
         md = addName(h);
     } else if (isCommandHandlerMetadata(h)) {
-        md = addName(h);
+        md = addName(h); // JESS: this seems mighty incomplete to me.
```
ddgenome: A note for later?

src/internal/invoker/resourceRecovery.ts
```diff
@@ -0,0 +1,35 @@
...
+                    .then(() => f2.how()
+                        .catch(error => {
+                            logger.warn("Failed to release resource %s: %s", f2.what, error);
+                        }))
+                , what: f1.what + " and " + f1.what,
+            };
+        }
+
+        const releaseEverything = ctx.releaseSteps
+            .reduce(both, { how: () => Promise.resolve(), what: "inconceivable" });
```
ddgenome: Does a simple `map` not work nicely with promises?

src/server/BuildableAutomationServer.ts
```diff
@@ -214,13 +214,8 @@ export class BuildableAutomationServer extends AbstractAutomationServer {
...
 
     private enrichContext(ctx: HandlerContext): HandlerContext {
-        return {
-            teamId: ctx.teamId,
-            correlationId: ctx.correlationId,
-            invocationId: ctx.invocationId,
-            messageClient: ctx.messageClient,
-            graphClient: ctx.graphClient ? ctx.graphClient : this.graphClient,
-        };
+        ctx.graphClient = ctx.graphClient || this.graphClient;
```
ddgenome: Before, you were creating a new object, now you are returning the same, possibly modified object. Is this distinction important here?

src/internal/invoker/resourceRecovery.ts
ddgenome: Looking at the tests, it appears this allows for continuing should something fail. Is that correct?

jessitron: yes, that is the goal. and it puts them in sequence... although I'm not certain that's necessary, it is kinda reassuring

src/server/BuildableAutomationServer.ts
jessitron: extremely. the context is now mutated and carries information back to the automation.
hmm, which brings up a point. I could put the mutated data in the invoking function, and instead pass a function inside the context which releases it. That would be safer. Then the context would close over some mutable data, but not hold it. Is that better?

src/internal/metadata/metadataReading.ts
jessitron: yeah, I suspect that whole case doesn't actually work. which is not related to this PR.

src/server/BuildableAutomationServer.ts
ddgenome: I don't know.

ddgenome:
Closing in favor of #136, which incorporates these changes in a slightly different way.

PR #134, opened 2017-11-28
Separate ignored changes from other changes

jessitron:
because usually when we ask whether a project is clean we mean, is there something to commit?
I was getting push errors in linting because the only change was in node_modules

this makes status.isClean ignore ignored files, while still storing the ignored file changes, so that my tests can check for whether it's _really_ clean. (also includes the raw output, would have helped with debugging this one)

This is consistent with GitProject.isClean(), which shall now be deprecated.


jessitron:
@ddgenome fyi, the ignored-files part is useful when I'm testing that a clone was fully cleaned out. I don't want you to update a node_module and that to sit around for the next person who uses the directory.

PR #136, opened 2017-11-29
expose disposable functionality on HandlerContext

cdupuis:
This is building on @jessitron's PR #129 but starting to expose lifecycle of `HandlerContext` on the context itself. IMHO this unifies the programming model and API of `HandlerContext`.

Let me know what you think.


src/internal/invoker/disposable.ts
```diff
@@ -0,0 +1,47 @@
+import { HandlerContext } from "../../HandlerContext";
+import { logger } from "../util/logger";
+
+export function registerDisposable(ctx: HandlerContext): (callback: () => Promise<void>, description?: string) => void {
```
ddgenome: Why return a function that adds the disposable instead of just adding the disposable?

src/internal/invoker/disposable.ts
```diff
@@ -0,0 +1,47 @@
...
+        if ((ctx as any).__disposables) {
+            (ctx as any).__disposables.push({ how: callback, what: description });
+        } else {
+            const disposables = [{ how: callback, what: description }];
+            (ctx as any).__disposables = disposables;
+        }
+    };
+}
+
+export function dispose(ctx: HandlerContext): () => Promise<void> {
```
ddgenome: Same here, why return a function? Is it because you imagine the common use case will be `.then(dispose)`?

src/internal/transport/AbstractRequestProcessor.ts
```diff
@@ -54,14 +59,18 @@ export abstract class AbstractRequestProcessor implements RequestProcessor {
                 mappedParameters: command.mapped_parameters,
                 secrets: command.secrets,
             };
-            const ctx: HandlerContext & AutomationContextAware = {
+            const ctx: any = {
```
ddgenome: Why detype here?

src/internal/transport/AbstractRequestProcessor.ts
```diff
@@ -89,14 +98,18 @@ export abstract class AbstractRequestProcessor implements RequestProcessor {
                 },
                 secrets: event.secrets,
             };
-            const ctx: HandlerContext & AutomationContextAware = {
+            const ctx: any = {
```
ddgenome: Same question, is it because AutomationContextAware has no lifecycle? Should it?

src/server/BuildableAutomationServer.ts
```diff
@@ -214,13 +214,8 @@ export class BuildableAutomationServer extends AbstractAutomationServer {
...
     private enrichContext(ctx: HandlerContext): HandlerContext {
-        return {
-            teamId: ctx.teamId,
-            correlationId: ctx.correlationId,
-            invocationId: ctx.invocationId,
-            messageClient: ctx.messageClient,
-            graphClient: ctx.graphClient ? ctx.graphClient : this.graphClient,
-        };
+        ctx.graphClient = ctx.graphClient || this.graphClient;
+        return ctx;
```
ddgenome: Same question I had on @jessitron's PR: is it safe to change the semantics here to modify the passed object instead of creating a new one? Plus, now that you've added the lifecycle property, should we be dealing with that here too?

src/internal/invoker/disposable.ts
cdupuis: That functions is being used when creating the context and gets assigned to `context.lifecycle.registerDisposable`. 

Since we don't have an actual implementation of `HandlerContext` that is way to park the implementation add it into the `HandlerContext` implementation when we create it.

src/internal/invoker/disposable.ts
cdupuis: That functions is being used when creating the context and gets assigned to `context.lifecycle.dispose`.

src/internal/transport/AbstractRequestProcessor.ts
cdupuis: Actually you're right, I don't think that is need any longer. I can change that back.

src/internal/transport/AbstractRequestProcessor.ts
cdupuis: Same as above.

src/server/BuildableAutomationServer.ts
cdupuis: @jessitron's implementation is correct. The old implementation already lost the `context` attribute. With her change we are safe. That's good.

jessitronica:
@cdupuis thanks for taking this and adjusting it to something that fits better. I like the things you changed.
(sorry I didn't get to it earlier, my schedule is odd today)

PR #140, opened 2017-11-30
Report metrics on clone caching

jessitron:
This will report about cached clone hits and misses every 10 minutes, both in total and per repo.

I suspect you might want to change the implementation of the "all day" part, the logCachingDirectoryManagerMetricsAllDay method.

### DIFF
--- a/src/spi/clone/CachingDirectoryManager.ts
+++ b/src/spi/clone/CachingDirectoryManager.ts
@@ -27,13 +27,14 @@ const cache = new StableDirectoryManager({
  * @type {{directoryFor:
  * ((owner: string, repo: string, branch: string, opts: CloneOptions) => Promise<CloneDirectoryInfo>)}}
  */
-export const CachingDirectoryManager: DirectoryManager = {
+export const CachingDirectoryManager: DirectoryManager & CachingOfClonesMetricsReporting = {
 
     directoryFor(owner: string, repo: string, branch: string, opts: CloneOptions): Promise<CloneDirectoryInfo> {
 
         return cache.directoryFor(owner, repo, branch, opts).then(existing =>
             pleaseLock(existing.path).then(lockResult => {
                 if (lockResult.success) {
+                    CachingOfClonesMetricsCacheImpl.incrementReuses({ owner, repo });
                     return {
                         ...existing,
                         release: () => {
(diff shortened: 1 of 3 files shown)


src/spi/clone/CachingDirectoryManager.ts
```diff
@@ -60,13 +67,116 @@ export const CachingDirectoryManager: DirectoryManager = {
...
+    promisify((a, b) => setTimeout(b, a));
+
+const MetricLogFrequency = 1000 * 60 * 10; // ten minutes
+
+/* This returns a promise that never finishes, unless it throws somehow
+ */
+export function logCachingDirectoryManagerMetricsAllDay(): Promise<void> {
+    return sleepPlease(MetricLogFrequency)
+        .then(() => logger.info(CachingOfClonesMetricsCacheImpl.report().print()))
+        .then(() => logCachingDirectoryManagerMetricsAllDay());
```
ddgenome: Is this tail recursion cool stack-wise?

src/spi/clone/CachingDirectoryManager.ts
```diff
@@ -60,13 +67,116 @@ export const CachingDirectoryManager: DirectoryManager = {
...
+            this.fallbackCounts[keyFor(repoId)] = 1;
+            return;
+        }
+        this.fallbackCounts[keyFor(repoId)]++;
+    }
+
+}
+
+function keyFor(repoId: RepoId) {
+    return `${repoId.owner}/${repoId.repo}`;
```
ddgenome: What should we do about owner/repo name collisions across GHE instances?

src/spi/clone/CachingDirectoryManager.ts
jessitron: the promise returns immediately, it stores the then internally. so it isn't recursive on the same promise thread.

I think so.

src/spi/clone/CachingDirectoryManager.ts
jessitron: I agree that's a thing and we're not handling it -- we're not handling GHE in tons of the client code.

cdupuis:
@jessitron, please see #142 for an integration of caching stats with the metrics support already in the client.

PR #143, opened 2017-11-30
Remove old hierarchy for editors, generators and r…

johnsonr:
The new functional style, as in `generatorToCommand`, is superior.

This is a breaking change, but does not affect newer automations in `spring-automation` and `node-automation`, which should be the model for other users and docs.


jessitron:

* 4 successful checks

PR #147, opened 2017-12-01
Fix clone token, unify repo creation, move cache

### LINKED ISSUE #63: The repository cache should be under ~/.atomist
Currently, the cache of checked out repos is `~/.atomist-working`, it should be under the `~/.atomist/` directory we already use rather than creating another top-level directory in people's home directory. We should also add a directory for provider to avoid conflicts of same-named repos in different providers. I propose `~/.atomist/repos/PROVIDER/OWNER/REPO`.

ddgenome:
Unify repo creation and move it to gitHub.ts.  Centralize creation of
the clone URL and ensure the token in the clone URL is always viewed
as such by GitHub.

Create directories with fs-extra.ensureDir, it is a lot easier and
more reliable.

Move repo cache to ~/.atomist/cache/repos/github.com/OWNER/REPO.

Closes #63

### DIFF
--- a/src/operations/common/RepoId.ts
+++ b/src/operations/common/RepoId.ts
@@ -13,7 +13,7 @@ export interface RepoId {
 
 export class SimpleRepoId implements RepoId {
 
-    constructor(public owner: string, public repo: string) {}
+    constructor(public owner: string, public repo: string) { }
 }
 
 /**
@@ -82,7 +82,7 @@ export class RemoteRepoRefSupport implements RemoteRepoRef {
     }
 
     public cloneUrl(creds: ProjectOperationCredentials) {
-        return `https://${creds.token}@${this.remoteBase}/${this.pathComponent}.git`;
+        return `https://${creds.token}:x-oauth-basic@${this.remoteBase}/${this.pathComponent}.git`;
     }
 
     get pathComponent(): string {
--- a/src/util/gitHub.ts
+++ b/src/util/gitHub.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosPromise } from "axios";
+import axios, { AxiosPromise, AxiosRequestConfig } from "axios";
 import { logger } from "../internal/util/logger";
 
 import { decode } from "../internal/util/base64";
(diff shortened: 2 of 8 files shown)


src/operations/common/RepoId.ts
```diff
@@ -82,7 +82,7 @@ export class RemoteRepoRefSupport implements RemoteRepoRef {
     }
 
     public cloneUrl(creds: ProjectOperationCredentials) {
-        return `https://${creds.token}@${this.remoteBase}/${this.pathComponent}.git`;
+        return `https://${creds.token}:x-oauth-basic@${this.remoteBase}/${this.pathComponent}.git`;
```
jessitronica: this is wonderful. Does it mean if I pass garbage as a token, it won't ask me for that user's password?

src/util/gitHub.ts
```diff
@@ -99,11 +99,32 @@ export function createCommitComment(token: string, rr: GitHubRepoRef, comment: C
...
 
-function authHeaders(token: string) {
+export function createRepo(token: string, rr: GitHubRepoRef, description: string, priv: boolean): AxiosPromise {
+    const config = authHeaders(token);
+    return axios.get(`${rr.apiBase}/orgs/${rr.owner}`, config)
+        .then(result => {
+            // We now know the owner is an org
+            return `${rr.apiBase}/orgs/${rr.owner}/repos`;
+        }, err => {
+            // We now know the owner is an user
```
jessitronica: I still think this is bad -- there are other errors besides this one. However, this is not part of the PR, it's just been there a while.

cdupuis: why is it bad?

jessitron: "the owner is a user" is not the only reason the call can fail

src/operations/common/RepoId.ts
ddgenome: Correct. That's the issue I ran into and was seeing that test fail. Hence the fix. Pulling that thread let to a couple other fixes.

PR #159, opened 2017-12-04
add generator benchmark test

claymccoy:
It can be run with "npm run benchmark".
There are options to change the number of generator executions, number of files, and amount of file content.
No GitHub usage or tokens are needed. No shell cwd monkeying is required.
Generator source is an in memory project and the target is a local file based project.
I originally wanted to test GenericGenerator, but it does not return a useful result from the generator call that it makes, and it doesn't do anything extra that is relevant. So we are testing the generate method directly so that we can get a useful result back that will tell us the location of the tmp directory where the generated project is.
After the first generator execution we run a basic assert to sanity check that everything worked right, this does not happen on the rest of the executions which should work the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
   "scripts": {
     "autostart": "npm-run-all --print-label --parallel watch:compile watch:server",
     "autotest": "supervisor --watch src,test --extensions ts --no-restart-on exit --quiet --exec npm -- test",
+    "benchmark": "mocha --prof --compilers ts:espower-typescript/guess \"test-perf/**/*Benchmark.ts\"; node --prof-process isolate-* > profile.txt",
     "build": "npm-run-all lint compile test test:api",
     "clean": "npm-run-all clean:js clean:build",
     "clean:build": "rimraf *-v8.log profile.txt build",
--- a/package.json
+++ b/package.json
@@ -117,7 +117,6 @@
   "scripts": {
     "autostart": "npm-run-all --print-label --parallel watch:compile watch:server",
     "autotest": "supervisor --watch src,test --extensions ts --no-restart-on exit --quiet --exec npm -- test",
-    "benchmark": "mocha --prof --compilers ts:espower-typescript/guess \"test-perf/**/*Benchmark.ts\"; node --prof-process isolate-* > profile.txt",
     "build": "npm-run-all lint compile test test:api",
     "clean": "npm-run-all clean:js clean:build",
     "clean:build": "rimraf *-v8.log profile.txt build",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -117,6 +117,7 @@
   "scripts": {
     "autostart": "npm-run-all --print-label --parallel watch:compile watch:server",
     "autotest": "supervisor --watch src,test --extensions ts --no-restart-on exit --quiet --exec npm -- test",
+    "benchmark": "mocha --prof --compilers ts:espower-typescript/guess \"test-perf/**/*Benchmark.ts\"; node --prof-process isolate-* > profile.txt",
```
ddgenome: It would be good to "scope" this under "test", i.e., "test:benchmark".

ddgenome: Also, since the benchmarks are under `test-perf`, is it safe to simplify the glob pattern to just `*.ts`?

test-perf/GeneratorBenchmark.ts
```diff
@@ -0,0 +1,86 @@
...
+        const repoName = "perfTestRepo";
+
+        const seed = constructProject(numberOfSeedProjectFiles, numberOfLipsumInFile);
+        generateFromSeed(seed, repoName)
+        .then(r => {
+            const result = r as ActionResult<LocalProject>;
+            NodeFsLocalProject.fromExistingDirectory(new GitHubRepoRef("atomist", repoName), result.target.baseDir)
+            .then(created => {
+                created.totalFileCount().then(fileCount =>
+                    assert.deepEqual(fileCount, numberOfSeedProjectFiles),
```
ddgenome: Is `deepEqual` necessary here?

test-perf/GeneratorBenchmark.ts
```diff
@@ -0,0 +1,86 @@
...
+                ).then(
+                    () => {
+                        let lastGeneration: Promise<any>;
+                        for (let i = 1; i < numberOfTimesToGenerate ; i++) {
+                            lastGeneration = generateFromSeed(seed, repoName);
+                        }
+                        if (lastGeneration) {
+                            lastGeneration.then(() => done());
```
ddgenome: Should this be using a `Promise.all` approach? The last generation could finish before all the rest, right?

package.json
claymccoy: sure, I thought about doing those things but I was just following the template from microgrammars

test-perf/GeneratorBenchmark.ts
claymccoy: yes, I think it is always necessary.

test-perf/GeneratorBenchmark.ts
claymccoy: probably in real production code where it really matters. It even works fine if done happens immediately after the assertion, it just looks weird because the test is still going.

ddgenome: That should probably be fixed then, right?

test-perf/GeneratorBenchmark.ts
ddgenome: I think it is required because the package.json `"directory": { "test": "test" }` means espower-typescript does not look outside the test directory, and this test resides outside it.

PR #171, opened 2017-12-08
Reduce tmp directory retention

ddgenome:


jessitron:
what does this do?
it looks like a lot of test changes, most (all?) of which are test cleanup/formatting

and then an unsafeCleanup flag?

possibly related: soon we'll be able to explicitly delete temp directories for clones after each invocation exits

ddgenome:
I noticed when testing that the temp directories were not being deleted. Where possible, this deletes them, and fixes the logic of a few tests as a bonus. Since the directories created by tempProject are cannot be deleted there, those still hang around. So this does not eliminate all created temp directories, hence "reduce"in the title.

PR #172, opened 2017-12-08
Wire up addAtomistWebhook

ddgenome:
Actually call addAtomistWebhook in the generator handler.


ddgenome:
Sorry if it was hard to find, but this `.then` https://github.com/atomist/automation-client-ts/pull/172/files#diff-258422caede6fece8f40f29b27a04a48R103 and the test is all new.

PR #189, opened 2017-12-31
fewer user/email calls

jessitron:
i want to have the ability to pass in a user/email, like from the graph.

The real problem here happens when someone is running the automation locally and it's using their local token which they generated themselves and didn't check "user/emails" permission. If we can pass in the name/email, then we can attribute the commit correctly even without having the token for the user who triggered the editor.


ddgenome:
Feel free to re-open when you have time to address the review and resolve conflicts.

PR #191, opened 2018-01-02
Create commit on existing branch

jessitron:
This lets a BranchCommit editMode add a commit to a branch that exists.

Rod, do you think this will mess anything up? It does change the default behavior in the case of branch-exists from "fail" to "push to that branch." I don't know of any current cases where that's not good.


jessitron:
if you check out bar, then it won't commit to foo. Checking out bar will have a different effect.

PR #200, opened 2018-01-10
Add gql-gen subcommand to atomist CLI

### LINKED ISSUE #130: Provide gql:gen script that doesn't fail if `graphql` does not exist
Right now the gql:gen script fails when there are no graphql files matching the pattern. We should fix that. 

That command should be wrapped with a CLI command, like `atomist generate typings` or so.

ddgenome:
Current approach in package script fails if there are no GraphQL
files.  This implementation simply wraps that with a check to see if
the glob pattern returns any files.  If it does, it passed the glob
along.  If not, it does not.

Closes #130


jessitron:

* 4 successful checks

PR #214, opened 2018-02-03
Subscription parameters

cdupuis:
This allows to specify the usage of parameters when using `GraqhQL.subscriptionFromFile(path, current, parameters))`

eg. the following graphql file has three parameters:

```
subscription SomeSubscription {
    ChatTeam(id: $teamId) {
        orgs {
            repo(first: 100, offset: $offset, private: $isPrivate) {
                owner
                name
            }
        }
    }
}
```

When using this with

```
const query = GraphQL.subscriptionFromFile(
            "./someSubscription",
            __dirname,
            {
                teamId: "T1L0VDKJP",
                offset: 100,
                isPrivate: true,
            });
```

the variables get replaced in the predicates.

NOTE: this doesn't give you full support for variables in subscriptions right now but we don't offer that in the API anyway as of today. This is a good first step.

### DIFF
--- a/src/graph/ApolloGraphClient.ts
+++ b/src/graph/ApolloGraphClient.ts
@@ -60,7 +60,8 @@ export class ApolloGraphClient implements GraphClient {
                                       variables?: Q,
                                       options?: any,
                                       current?: string): Promise<T> {
-        return this.executeQuery<T, Q>(resolveAndReadFileSync(queryFile, current), variables, options);
+        return this.executeQuery<T, Q>(
+            resolveAndReadFileSync(queryFile, current, {}), variables, options);
     }
 
     public executeQuery<T, Q>(graphql: string,
@@ -77,10 +78,10 @@ export class ApolloGraphClient implements GraphClient {
         });
 
         return this.client.query<T>({
-            query,
-            variables,
-            ...options,
-        })
+                query,
+                variables,
+                ...options,
+            })
             .then(result => callback(result));
     }
 
@@ -88,7 +89,8 @@ export class ApolloGraphClient implements GraphClient {
                                          variables?: Q,
                                          options?: any,
                                          current?: string): Promise<T> {
-        return this.executeMutation<T, Q>(resolveAndReadFileSync(mutationFile, current), variables, options);
+        return this.executeMutation<T, Q>(
+            resolveAndReadFileSync(mutationFile, current, {}), variables, options);
     }
 
     public executeMutation<T, Q>(graphql: string,
@@ -106,10 +108,10 @@ export class ApolloGraphClient implements GraphClient {
         });
 
         return this.client.mutate<T>({
-            mutation,
-            variables,
-            ...options,
-        })
+                mutation,
+                variables,
+                ...options,
+            })
             .then(response => callback(response));
     }
 
--- a/src/graph/graphQL.ts
+++ b/src/graph/graphQL.ts
@@ -18,11 +18,16 @@ const schema = require("./schema.cortex.json");
  * Note: Use __dirname to get the current directory of the calling script.
  * @param {string} path
  * @param {string} current
+ * @param {{[p: string]: string | boolean | number}} parameters
  * @returns {string}
  */
-export function subscriptionFromFile(path: string, current: string = appRoot.path): string {
+export function subscriptionFromFile(path: string,
+                                     current: string = appRoot.path,
+                                     parameters: {
+                                        [name: string]: string | boolean | number;
+                                     } = {}): string {
     // TODO cd add validation that we only read subscriptions here
-    return resolveAndReadFileSync(path, current);
+    return resolveAndReadFileSync(path, current, parameters);
 }
 
 /**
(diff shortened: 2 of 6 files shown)


src/graph/ApolloGraphClient.ts
```diff
@@ -60,7 +60,8 @@ export class ApolloGraphClient implements GraphClient {
                                       variables?: Q,
                                       options?: any,
                                       current?: string): Promise<T> {
-        return this.executeQuery<T, Q>(resolveAndReadFileSync(queryFile, current), variables, options);
+        return this.executeQuery<T, Q>(
+            resolveAndReadFileSync(queryFile, current, {}), variables, options);
```
ddgenome: Perhaps parameters should be an optional argument so you don't have to add passing `{}` everywhere.

src/graph/graphQL.ts
```diff
@@ -18,11 +18,16 @@ const schema = require("./schema.cortex.json");
...
  * @param {string} current
+ * @param {{[p: string]: string | boolean | number}} parameters
  * @returns {string}
  */
-export function subscriptionFromFile(path: string, current: string = appRoot.path): string {
+export function subscriptionFromFile(path: string,
+                                     current: string = appRoot.path,
+                                     parameters: {
+                                        [name: string]: string | boolean | number;
+                                     } = {}): string {
```
ddgenome: Or perhaps this is the actual function that is called everywhere?

src/graph/graphQL.ts
```diff
@@ -72,15 +77,34 @@ export function prettyPrintErrors(errors: GraphQLError[], query?: string): strin
...
-        return fs.readFileSync(absolutePath).toString();
+        let content = fs.readFileSync(absolutePath).toString();
+        for (const key in parameters) {
+            if (parameters.hasOwnProperty(key)) {
+                const value: any = parameters[key];
+                if (typeof value === "string") {
+                    content = content.replace(`$${key}`, `"${value}"`);
+                } else if (typeof value === "number") {
+                    content = content.replace(`$${key}`, `${value}`);
+                } else if (typeof value === "boolean") {
```
ddgenome: Seems the "number" and "boolean" cases could be combined into one. Or you could get rid of the conditional entirely with something like:

```typescript
content = content.replace(new RegExp(`\\$${key}\\b`, "g"), JSON.stringify(parameters[key]));
```

test/graph/GraphQLTest.ts
```diff
@@ -58,4 +58,18 @@ query Repos($teamId: ID!, $offset: Int!) {
...
+        const query = GraphQL.subscriptionFromFile(
+            "./someSubscription",
+            __dirname,
+            {
+                teamId: "T1L0VDKJP",
+                offset: 100,
+                isPrivate: true,
+            });
+        assert(query.includes(`"T1L0VDKJP"`));
+        assert(query.includes("100"));
```
ddgenome: This will always be true because 100 already appears in the subscription before replacing the variables.

test/graph/GraphQLTest.ts
```diff
@@ -58,4 +58,18 @@ query Repos($teamId: ID!, $offset: Int!) {
...
+        assert(query.includes("private: true"));
```
ddgenome: The query is small enough that I'd have the properly replaced version as a string and assert ===.

src/graph/graphQL.ts
```diff
@@ -72,15 +77,34 @@ export function prettyPrintErrors(errors: GraphQLError[], query?: string): strin
...
     }
     const absolutePath = p.resolve(current, path);
     if (fs.existsSync(absolutePath)) {
-        return fs.readFileSync(absolutePath).toString();
+        let content = fs.readFileSync(absolutePath).toString();
+        for (const key in parameters) {
+            if (parameters.hasOwnProperty(key)) {
+                const value: any = parameters[key];
+                if (typeof value === "string") {
+                    content = content.replace(`$${key}`, `"${value}"`);
```
ddgenome: This will only replace the first occurrence of `$VARIABLE`.

ddgenome: And will also match the variable name as a substring. For example, if you had a variable named `$foo` and a second named `$fooBar`, performing the replacement on `$foo` will match `$fooBar` should that come first in the string.

test/graph/GraphQLTest.ts
ddgenome: It's probably worth adding tests for multiple replacements of a single variable and the `$foo`/`$fooBar` case.

test/graph/someSubscription.graphql
```diff
@@ -0,0 +1,10 @@
+query SomeSubscription {
```
ddgenome: Aren't you expected to list the variables here too?

```
query SomeSubscription($teamId: string, $offset: number, $isPrivate: boolean) { ...
```

I guess that would make replacement harder though, since that would have to be removed since we aren't actually passing them along to the GraphQL parser.

http://graphql.org/learn/queries/#variables

src/graph/ApolloGraphClient.ts
cdupuis: Good point. Adding.

src/graph/graphQL.ts
cdupuis: Yeah, that is the only function we call.

src/graph/graphQL.ts
cdupuis: Odd, you're right although
```
/**
      * Replaces text in a string, using a regular expression or search string.
      * @param searchValue A string to search for.
      * @param replaceValue A string containing the text to replace for every successful match of searchValue in this string.
      */
```

src/graph/graphQL.ts
cdupuis: Thanks, that's good.

test/graph/GraphQLTest.ts
cdupuis: True

test/graph/someSubscription.graphql
cdupuis: That's the part that is tricky. I couldn't find a way to use the parser to remove those. I've added a regex match and remove the match now. Take a look.

test/graph/GraphQLTest.ts
cdupuis: Maybe you can come up with some regexp trickery. But for now I'm not too concerned about that problem.

slimslenderslacks:

* 5 successful checks

Issue #220, opened 2018-02-21
Returning undefined from a handler causes a TypeError


jessitron:
since it can return `any`, I think we should treat undefined as success? ... or not a TypeError, anyway

PR #224, opened 2018-02-22
Add a test that shows a failure

jessitron:
It throws an exception and also makes an unhandled promise exception

@cdupuis this is what I meant by #220 -- your resolution of that one showed me where to write the test. Now you can see the TypeError that I got from our automation


cdupuis:
Oh, that's great. Obviously I didn't read #220 correctly. I'll fix this asap.

jessitron:
fixed in master, thanks

PR #237, opened 2018-03-15
GitProject always has a remote repo ref

jessitron:
Declaring it lets us access its URL


jessitron:
forget it. it's too hard

PR #239, opened 2018-03-15
Print a fancy message on registration

jessitron:
"Websocket connected" tells us what we need, but it isn't obvious to other people


cdupuis:
I'll take that and probably suggest some changes. Thanks @jessitron.

jessitron:
i figured you would! you usually build on my PRs

PR #240, opened 2018-03-16
Centralize configuration handling into client

ddgenome:
Currently it is difficult to use the capabilities available to you in
container runtimes to override the automation configuration.  This
organizes configuration interfaces, renames some with backward
compatibility, and defines a clear hierarchy of configuration sources
that allow for straightforward use of container runtime options like
Kubernetes secrets and Cloud Foundry services.

This also allows the user and module configs to specify pretty much
any automation config value, not just team IDs and token.

Add TypeDoc for configuration interfaces.

These changes are largely backward-compatible.  The exceptions are the
location of the server option interfaces, which most likely are not
referenced outside this package, and restricting teamIds to a string[]
rather than either a string or string[].  The latter will result in a
compile error in clients that use a string and is an easy fix.  The
changes allow automationClient.ts to be simplified.

Also, fix --version command-line option.


src/cli/commands.ts
```diff
@@ -32,10 +32,16 @@ export function extractArgs(args: string[]): Arg[] {
 
 export function readVersion(): string {
     try {
-        const pj = require(`${appRoot}/package.json`);
+        const pj = require(p.join(__dirname, "..", "package.json"));
         return `${pj.name} ${pj.version}`;
     } catch (e) {
-        return "@atomist/automation-client 0.0.0";
+        try {
+            // in case you are running locally
```
jessitron: yay!

src/configuration.ts
```diff
@@ -1,180 +1,644 @@
...
+    /**
+     * Groups this automation will be registered with.  Must be
+     * specified if teams is not specified.  Cannot be specified if
+     * teams is specified.
+     */
+    groups?: string[];
+    /**
+     * If events should be queued when the registration is not
+     * connected to the websocket, specificy "durable".  "ephemeral"
+     * is suited for testing and running locally.
```
jessitron: I bet "ephemeral" is the default

src/configuration.ts
```diff
@@ -1,180 +1,644 @@
...
+     */
+    policy?: "ephemeral" | "durable";
+    /**
+     * GitHub personal access token with repo and read:org scopes for
+     * all GitHub users and organizations associated with the Atomist
+     * teams this automation is registered in.
+     */
```
jessitron: aaaah this documentation is so beautiful

what if someone else associates their user? can no one else register an automation? (this is not a problem with this change, this stuff is just confusing)

src/configuration.ts
```diff
@@ -1,180 +1,644 @@
...
+ * ATOMIST_TEAMS environment variable takes precedence over the
+ * ATOMIST_TEAM environment variable, which takes precedence over the
+ * configuration "teamdIds", which takes precedence over cfg.teamIds,
+ * which may be undefined, null, or an empty array.
+ */
+export function resolveTeamIds(cfg: Configuration): string[] {
+    if (process.env.ATOMIST_TEAMS) {
+        cfg.teamIds = process.env.ATOMIST_TEAMS.split(",");
+    } else if (process.env.ATOMIST_TEAM) {
+        cfg.teamIds = [process.env.ATOMIST_TEAM];
```
jessitron: nice

test/configurationTest.ts
```diff
@@ -58,34 +125,540 @@ describe("configuration", () => {
...
+            assert.deepStrictEqual(c, defCfg);
+        });
+
+    });
+
+    describe("loadUserConfiguration", () => {
+
+        it("should warn but return original", () => {
+            const save = process.env.HOME;
+            process.env.HOME = "/this/file/should/not/exist/so/please/do/not/make/it";
```
jessitron: tempting

jessitron:
wow the comments in here are beautiful. Thank you for putting all the thought into this.

src/configuration.ts
ddgenome: I see how this is confusing, particularly the user part.

ddgenome:
Superseded by #241. Comments here addressed there.

PR #241, opened 2018-03-16
Additional async initialisers

cdupuis:
Some configuration values can only be obtained via async calls. Added `initializers` that allows to register functions that return Promise. These promises are being resolved during loading of the configuration at the very end.

This allows us to hook in things like axios async http calls to obtain secrets or similar.

@ddgenome I added this because I wasn't able to see how else we would be able to remove the custom start scripts etc. Those start scripts use axios to obtain secrets and I didn't see a way to get those Promises into the configuration.custom thingy. Probably there is a better solution to this in which case we can just close this PR.


test/configurationTest.ts
```diff
@@ -58,34 +126,560 @@ describe("configuration", () => {
...
+
+                    _.forEach(save, (v, k) => {
+                        if (v) {
+                            process.env[k] = v;
+                        } else {
+                            delete process.env[k];
+                        }
+                    });
+                    done();
+                });
```
ddgenome: Instead of the `done()` in the line above, there should be a `.then(() => done(), done)` at the end. Or, since we can use async/await now, we could do that.

src/configuration.ts
```diff
@@ -1,180 +1,674 @@
...
 /**
- * Read and return user config.
+ * Read and return user config from UserConfigFile.
  */
 export function getUserConfig(): UserConfig {
-    try {
-        if (fs.existsSync(UserConfigFile)) {
-            return fs.readJsonSync(UserConfigFile) as UserConfig;
+    let cfg: UserConfig;
+    if (fs.existsSync(userConfigPath())) {
```
ddgenome: If the file is not there the try will properly handle and report that case, as well as if it is not readable, not proper JSON, etc. It also avoids a race condition.

From https://nodejs.org/api/fs.html#fs_fs_exists_path_callback:

> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

cdupuis: I want to avoid even attempting to load the file and not report the fact it is missing. IMHO it looks bad if the first thing we log when the client starts is a warning about a missing file. 

That’s why I think using exists to check if the file is there is ok. If we ever hit a race condition that suddenly the file goes missing, this is still being handled in the inner catch.

test/configurationTest.ts
cdupuis: Yeah, that’s right. But wouldn’t `.then(done, done)` also work?

ddgenome:
@cdupuis I've made all the changes we discussed so this is OK to merge from my perspective.

ddgenome:
When I update a client to use this pre-release version, mocha tests in the clients all run successfully, but the process never exits. Any ideas why?

PR #245, opened 2018-03-18
Add branch as an optional field to RepoRef

jessitron:
We often want to specify both this and the SHA
especially for editors; they must have a branch. and statuses must have a sha.

### DIFF
--- a/src/operations/common/GitHubRepoRef.ts
+++ b/src/operations/common/GitHubRepoRef.ts
@@ -7,6 +7,7 @@ import axios from "axios";
 import { logger } from "../../internal/util/logger";
 import { Configurable } from "../../project/git/Configurable";
 import { AbstractRepoRef } from "./AbstractRemoteRepoRef";
+import { GitShaRegExp } from "./params/gitHubPatterns";
 
 export const GitHubDotComBase = "https://api.github.com";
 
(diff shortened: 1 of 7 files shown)


src/operations/common/GitHubRepoRef.ts
```diff
@@ -15,6 +16,18 @@ export const GitHubDotComBase = "https://api.github.com";
  */
 export class GitHubRepoRef extends AbstractRepoRef {
 
+    public static for(params: { owner: string, repo: string, sha?: string, rawApiBase?: string, path?: string, branch?: string }): GitHubRepoRef {
```
johnsonr: Not keen on using a reserved word as a function name even if it is legal

src/operations/support/editorUtils.ts
```diff
@@ -92,7 +92,7 @@ function doWithEditResult(r: EditResult<GitProject>, gitop: () => Promise<EditRe
                 }) : gitop();
             });
     }
-    logger.info("NOT committing %j as it's not dirty, edited=%s", r.target.id, r.edited);
+    logger.info("NOT committing %j in %s as it's not dirty, edited=%s", r.target.id, r.target.baseDir, r.edited);
```
johnsonr: I think we should add logging on all the branches in this function. It's important information and will be invaluable for debugging

src/operations/common/GitHubRepoRef.ts
jessitron: oh good call

src/operations/support/editorUtils.ts
jessitron: ok. yeah this came up for us as I was working on this so I threw it in

Issue #244, opened 2018-03-17
Support graphql fragments


jessitron:

jessitron:
Related issue atomisthq/automation-api#390 created

cdupuis:
Related issue atomisthq/neo4j-ingester#360 linked

PR #250, opened 2018-03-21
Initial support for fragments in queries and subscriptions

cdupuis:
This should address the basic use case of #244.


src/graph/ApolloGraphClient.ts
```diff
@@ -92,26 +113,41 @@ export class ApolloGraphClient implements GraphClient {
...
+        if (typeof options === "string") {
+            options = {
+                name: options,
+            };
+        }
+        const m = graphql.mutate({
+            mutation: options.mutation,
+            path: options.path,
+            name: options.name,
+            moduleDir: trace.get()[1].getFileName(),
```
jessitron: oh that is _sneaky_.

I like it, although we have to be careful to never wrap these functions internally -- I'd rather make that clear. let me put in a commit

src/graph/graphQL.ts
```diff
@@ -25,13 +29,185 @@ export function enumValue(value: string | string[]): ParameterEnum {
...
+ */
+export function query<T, Q>(options: QueryOptions): string {
+    let q = options.query;
+    const fragmentDir = options.fragmentDir;
+    const path = options.path;
+    const name = options.name;
+
+    // If query isn't defined attempt to load from file
+    if (!q) {
+        q = locateAndLoadGraphql({ path, name }, "query", options.moduleDir);
```
jessitron: this one does not autopopulate the module dir from the stack trace?

src/graph/graphQL.ts
```diff
@@ -25,13 +29,185 @@ export function enumValue(value: string | string[]): ParameterEnum {
...
+    }
+    return q;
+}
+
+export interface QueryOptions {
+    query?: string;
+    path?: string;
+    name?: string;
+    fragmentDir?: string;
+    moduleDir: string;
```
jessitron: moduleDir is required even if `query` is provided?

src/graph/graphQL.ts
```diff
@@ -116,40 +292,111 @@ export function resolveAndReadFileSync(path: string,
...
+    return q;
 }
 
-function replace(query: string, key: string, value: string): string {
-    return query.replace(new RegExp(`${key}\\b`, "g"), value);
+function replace(q: string, key: string, value: string): string {
+    return q.replace(new RegExp(`${key}\\b`, "g"), value);
+}
+
+function inlineFragments(q: string, name: string, moduleDir: string, fragmentDir: string): string {
```
jessitron: which of these are required? i am going to try to get this too

src/automationClient.ts
```diff
@@ -94,12 +94,12 @@ export class AutomationClient {
 
         setLogLevel(this.configuration.logging.level);
 
-        if (this.configuration.logging.file !== false) {
+        if (this.configuration.logging.file && this.configuration.logging.file.enabled === true) {
```
ddgenome: Is the first check necessary given the default configuration?

src/graph/graphQL.ts
```diff
@@ -116,40 +292,111 @@ export function resolveAndReadFileSync(path: string,
...
+
+    if (!fragmentDir && !name) {
+        fragmentDir = p.dirname(moduleDir);
+    } else if (!fragmentDir && name) {
+        fragmentDir = p.resolve(findUp.sync("graphql", { cwd: p.resolve(p.dirname(moduleDir)) }), "fragment");
+    } else if (!p.isAbsolute(fragmentDir)) {
+        fragmentDir = p.resolve(p.dirname(moduleDir), fragmentDir);
+    }
```
jessitron: is this intended to be consistent with the seeking of the graphql file itself? (not saying it isn't, suggesting we put that in a common function to express the consistency)

src/graph/graphQL.ts
```diff
@@ -116,40 +292,111 @@ export function resolveAndReadFileSync(path: string,
...
+        const graphqlDir = findUp.sync("graphql", { cwd: p.resolve(p.dirname(moduleDir)) });
+        if (graphqlDir) {
+            if (!name.endsWith(".graphql")) {
+                name = `${name}.graphql`;
+            }
+            path = p.resolve(graphqlDir, subfolder, name);
+        }
+    }
+
+    if (fs.existsSync(path)) {
```
ddgenome: You could do away with this test.

src/project/git/GitCommandGitProject.ts
```diff
@@ -195,6 +195,14 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
         return this.runCommandInCurrentWorkingDirectory(`git checkout ${sha} --`);
     }
 
+    /**
+     * Revert all changes since last commit
+     * @return {any}
+     */
+    public async revert(): Promise<CommandResult<this>> {
```
ddgenome: ???

src/graph/graphQL.ts
cdupuis: that's right, that method is really only used from within `ApolloGraphClient`. Ideally this would package protected. The moduleDir stuff is handled there.

test/graph/chatChannelFragment.graphql
```diff
@@ -0,0 +1,3 @@
+fragment ChannelName on ChatChannel {
```
ddgenome: Why is there a graph and a graphql directory under test?

src/graph/graphQL.ts
cdupuis: See above, this method is not meant to be used directly from clients. I wish I could hide it. This and `mutate` are only supposed to be used from `ApolloGraphClient`.

src/graph/graphQL.ts
cdupuis: What do you mean? They are all required parameters.

src/graph/graphQL.ts
cdupuis: I'm not sure I understand. Lets zoom.

src/automationClient.ts
cdupuis: yeah, I guess not. that's right. I'll remove it for clarity.

src/project/git/GitCommandGitProject.ts
cdupuis: Not sure what I did here but I ended up collection a bunch of things in this PR that were in master already.

test/graph/chatChannelFragment.graphql
cdupuis: Good question. That shouldn't be.

cdupuis: Ah, no that is fine: `graph` is the package that have the tests in and also some graphql files for testing relative imports. `graphql` is for testing resolution from a graphql directory.

PR #252, opened 2018-03-24
provide command to download/introspect team schema

cdupuis:
running `atomist gql-fetch <teamId> --token <token>` will now download and place the team schema in src/graphql.

`atomist gql-gen` will look for team specific schema at that location and use it if it is there. Will fall back to the schema in automation-client if a custom schema isn't available.


src/cli/commands.ts
```diff
@@ -67,15 +67,33 @@ export function run(
     return execNode("cli/run.js", args, msg, path, runInstall, runCompile);
 }
 
+export function gqlIntrospect(
+    path: string,
+    teamId: string,
+    token: string = process.env.GITHUB_TOKEN,
```
ddgenome: Should we also accept ATOMIST_TOKEN here?

src/start.cli.ts
```diff
@@ -85,6 +86,36 @@ yargs.completion("completion")
         }
 
     })
+    .command(["gql-introspect <team>"], "Introspect GraphQL schema", ya => {
```
ddgenome: While "introspect" is what apollo uses, it seems a bit abstract to me. Would "fetch" or something like that be more clear?

cdupuis:
Good point about `ATOMIS_TOKEN`. I've added that to the default as the first option. 

Once we have API keys we need to change the auth scheme to, but that's not something we can do now.

PR #268, opened 2018-04-04
Externalize default config values

cdupuis:
Configuration default values are now stored in the json files and are loaded based on `ATOMIST_ENV` or `NODE_ENV`. We do support the following mapping:

* no env -> `configuration.development.json`
* production -> `configuration.production.json`
* staging -> `configuration.staging.json`

I tried to come up with sensible values for those defaults based on environment but let's review.


src/configuration.development.json
```diff
@@ -0,0 +1,54 @@
+{
+  "teamIds": [],
+  "groups": [],
+  "environment": "local",
```
ddgenome: If the environment is "local", I'd recommend the file be named "configuration.local.json". Actually, I think using dashes is better than periods within the file name, so "configuration-local.json". I think "local" is more clear than "development".

src/configuration.production.json
```diff
@@ -0,0 +1,54 @@
...
+  "environment": "production",
+  "policy": "durable",
+  "endpoints": {
+    "api": "https://automation.atomist.com/registration",
+    "graphql": "https://automation.atomist.com/graphql/team"
+  },
+  "http": {
+    "enabled": true,
+    "host": "localhost",
+    "port": 2866,
```
ddgenome: There's a lot of duplication, might it not be better to keep the default and then override just what needs overriding with an additional `loadEnvironmentConfig()`?

ddgenome: That would also resolve the `null`/`undefined` issue.

src/configuration.staging.json
```diff
@@ -0,0 +1,54 @@
+{
+  "teamIds": [],
+  "groups": [],
+  "environment": "staging",
```
ddgenome: "testing" is much more common than "staging". Should we change the names here too are do you think this is mostly for us?

src/configuration.development.json
```diff
@@ -0,0 +1,54 @@
...
+    "file": {
+      "enabled": true,
+      "level": "debug"
+    },
+    "banner": true
+  },
+  "statsd": {
+    "enabled": false
+  },
+  "commands": null,
```
ddgenome: `null` and `undefined` are not treated the same by lodash, so this may prevent command and events from sources like atomist.config.js being merged in properly. There may not be a test for this.

ddgenome: Of course, the custom handling of array merging may do the right thing here.

ddgenome:
Those line comments are all out of order.

src/configuration.development.json
cdupuis: Yeah, that makes more sense. I'll change it.

src/configuration.development.json
cdupuis: I think I checked the null/undefined thing but maybe not. I'll check again.

src/configuration.production.json
cdupuis: Yeah, I think that would be cleaner. Let me try that and see how we like that.

src/configuration.staging.json
cdupuis: Right. yeah.

cdupuis: This is also where I initially wanted to support custom profile names that would be resolved from `appRoot.path`.

PR #275, opened 2018-04-17
Print the stack when it fails to load

jessitron:
this will help with diagnosing load errors. I get these a lot.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -475,6 +475,7 @@ export function loadAutomationConfig(cfgPath?: string): Configuration {
             cfg = require(cfgPath).configuration;
             cfgLog("automation config");
         } catch (e) {
+            logger.error(e.stack);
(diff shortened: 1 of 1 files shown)


src/configuration.ts
```diff
@@ -475,6 +475,7 @@ export function loadAutomationConfig(cfgPath?: string): Configuration {
             cfg = require(cfgPath).configuration;
             cfgLog("automation config");
         } catch (e) {
+            logger.error(e.stack);
```
ddgenome: Would it be better to modify the message of the existing Error, `e`, and rethrow that?

jessitron: I don't know. If we did that, we'd need to make whatever catches it and prints it show the stack trace.
This right here is a big improvement over not having the stack trace at all.
I hit this again now, and I really need it.

PR #278, opened 2018-04-26
subscribe to AtomistLog events

cdupuis:


src/internal/transport/OnLog.ts
```diff
@@ -0,0 +1,111 @@
+import { logger } from "../..";
```
ddgenome: I thought we didn't want to import from index.ts within the same project.

cdupuis: That's there so that other people can implement a `LogHandler` without the need to import from an internal directory.

cdupuis:
@ddgenome is that ok now?

PR #284, opened 2018-05-03
Reduce grace period, add namespace to app event

ddgenome:
Reduce default websocket grace period to 10 seconds.

Add namespace to the application event sent to Atomist.


src/internal/env/applicationEvent.ts
```diff
@@ -55,6 +55,7 @@ export function registerApplicationEvents(teamId: string): Promise<any> {
         pod: env ? env.instance_id : os.hostname(),
         host: env ? env.instance_id : os.hostname(),
         id: env ? env.instance_id : guid(),
+        namespace: env ? env.space_name : process.env.ATOMIST_ENV || process.env.NODE_ENV || "unknown",
```
cdupuis: The payload of the app event is defined in `incoming-webhooks`. I don't think we can just change it like that.

PR #285, opened 2018-05-04
Print a more specific error message

jessitron:
the previous one was "GraphQL file 'undefined' does not exist"

### DIFF
--- a/src/internal/graph/graphQL.ts
+++ b/src/internal/graph/graphQL.ts
@@ -309,13 +309,18 @@ function locateAndLoadGraphql(options: {
             path = p.resolve(p.dirname(moduleDir), path);
         }
     } else if (options.name) {
-        const graphqlDir = findUp.sync("graphql", { cwd: p.resolve(p.dirname(moduleDir)) });
+        const cwd = p.resolve(p.dirname(moduleDir));
+        const graphqlDir = findUp.sync("graphql", { cwd });
         if (graphqlDir) {
             if (!name.endsWith(".graphql")) {
                 name = `${name}.graphql`;
             }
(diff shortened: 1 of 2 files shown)


src/internal/graph/graphQL.ts
```diff
@@ -309,13 +309,18 @@ function locateAndLoadGraphql(options: {
...
         if (graphqlDir) {
             if (!name.endsWith(".graphql")) {
                 name = `${name}.graphql`;
             }
             path = p.resolve(graphqlDir, subfolder, name);
+        } else {
+            throw new Error("No graphql folder found anywhere above directory " + cwd + "\nConsider specifying a path");
         }
+    } else {
+        throw new Error("No name or path specified");
```
cdupuis: This is a good change.

src/internal/transport/websocket/WebSocketClient.ts
```diff
@@ -253,6 +254,18 @@ function register(registrationCallback: () => any, options: WebSocketClientOptio
...
+function warnAboutChatTeamId(registrationPayload: any) {
+    if (registrationPayload.team_ids.length === 1) {
+        const singleTeam = registrationPayload.team_ids[0];
+        if (singleTeam.startsWith("T")) {
+            logger.error(`This looks like a Chat Team: ${singleTeam}. You might want your Atomist Team ID, which usually starts with A.` +
+                "Try `@atomist pwd` in Slack to see both.");
+        }
+    }
+}
+
```
cdupuis: I don't think these kind of validation messages belong into the client. This should be handled in the API and the client should just display the error message. Also discussed this with @slimslenderslacks.

PR #286, opened 2018-05-04
Support for injection of configuration values

cdupuis:
This allows us to inject values from the configuration object into command and/or event handlers. 

As an example:
```
interface DockerOptions {
   registry: string;
   user: string;
   password: string;
}

... 

@Value({ path: "docker", root: "custom", required: false })
public dockerOptions: DockerOptions;
```

with the following configuration: 
```
export const configuration: Configuration = {
   custom: {
     docker: {
       registry: "http://google.com",
       user: "something",
       password: "something else",
     },
   },
};
```

If access to configuration is needed from outside command or event handlers, there is now the following method: 

```
const dockerOptions = configurationValue<DockerOptions>("custom.docker");
```

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -715,6 +715,10 @@ function validateConfiguration(cfg: Configuration) {
  * ATOMIST_TOKEN and ATOMIST_TEAMS environment variables,
  * respectively.
  *
+ * Placeholders in string config values will get resolved against the
+ * environment. The resolution happens at the very end when all configs
+ * have been merged.
+ *
  * The configuration exported from the atomist.config.js is modified
  * to contain the final configuration values and returned from this
  * function.
@@ -735,6 +739,7 @@ export function loadConfiguration(cfgPath?: string): Promise<Configuration> {
         resolveTeamIds(cfg);
         resolveToken(cfg);
         resolvePort(cfg);
+        resolvePlaceholders(cfg);
     } catch (e) {
         logger.error(`Failed to load configuration: ${e.message}`);
         if (e.stack) {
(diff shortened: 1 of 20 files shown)


ddgenome:
I can take a look at the changes in more detail later but on first glance this seems like a good idea. I'd ask you consider combining root and path into just path, making it a JSON path you can feed to _.get.

cdupuis:
That’s a good suggestion. I’ll change that back. It was what I had first. Also add defaults to @Value.

src/internal/metadata/decoratorSupport.ts
```diff
@@ -129,12 +129,47 @@ export function declareMappedParameter(target: any, name: string, uri: string, r
...
+        const found: any[] = params.filter(p => p.localKey === name);
+        if (found != null && found.length > 0) {
+            const index = params.indexOf(found[0]);
+            params.splice(index, 1);
+        }
+    }
+    const param = { name, path, required };
+    params.push(param);
+
+    // merge mapped_parameters from parent if it has some
```
jessitron: merge configuration values?

jessitron:
+1

cdupuis:
> Good idea. I assume you can also inject an object from the config?
@johnsonr, yes. I tried to show that in the PR description where DockerOptions get injected from the config. Is that not what you meant?

src/configuration.ts
```diff
@@ -307,6 +308,24 @@ export function defaultConfiguration(): Configuration {
...
+ * Exposes the configuration for lookup of configuration values.
+ * This is useful for components to obtain values eg. from configuration.custom
+ * like user provided secrets etc.
+ * @param {string} path the property path evaluated against the configuration instance
+ * @returns {T}
+ */
+export function configurationValue<T>(path: string, defaultValue?: T): T  {
+    const conf = automationClientInstance().configuration;
+    const value = _.get(conf, path) as T;
+    if (!value && !defaultValue) {
```
ddgenome: This might be more clear:

```
if (value) {
    return value;
} else if (defaultValue) {
    return defaultValue;
}
throw new Error(...);

File: src/internal/parameterPopulation.ts [hunk 3]
```diff
@@ -19,34 +21,46 @@ export function populateParameters(instanceToPopulate: any, hm: CommandHandlerMe
...
+    });
+}
+
+function computeValue(parameter: { name: string, type?: ParameterType }, value: any) {
     // Convert type if necessary
     switch (parameter.type) {
         case undefined:
             // It's a string. Keep the value the same
-            value = arg.value;
+            value = value;
```
ddgenome: Why are these `value = value` lines necessary?

test/atomist.config.ts
```diff
@@ -82,10 +82,10 @@ export const configuration: Configuration = {
         // workers: 2,
     },
     postProcessors: [
-        config => {
+        /*config => {
```
ddgenome: Should this still be commented out?

test/internal/metadata/classStyleMetadataReadingTest.ts
```diff
@@ -193,6 +286,9 @@ export class Superclass implements HandleCommand {
     @Secret("atomist://some_secret")
     public superSecret: string;
 
+    @Value("name")
+    public automationClientName: string;
```
ddgenome: Couldn't you just set the value in this case?

```
public automationClientName: string = "name";
```

ddgenome: Nevermind. I now understand this is referencing a path into the configuration.

src/configuration.ts
cdupuis: oh, yes.

src/internal/parameterPopulation.ts
cdupuis: no.

test/atomist.config.ts
cdupuis: yes

jrday:

* 8 successful checks

[atomist:generated] [atomist:auto-merge]

PR #294, opened 2018-05-18
Set branch after checking out

### LINKED ISSUE #293: GitCommandGitProject.checkout should set this.branch
If the checkout is successful, then `this.branch` should be set to the new branch.

ddgenome:
After successfully creating or checking out a branch, set the branch
property on GitCommandGitProject.

Closes #293

### DIFF
--- a/src/project/git/GitCommandGitProject.ts
+++ b/src/project/git/GitCommandGitProject.ts
@@ -176,7 +176,7 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
      * @param body
      */
     public raisePullRequest(title: string, body: string = name, targetBranch: string = "master"): Promise<ActionResult<this>> {
-        if (!(this.branch)) {
+        if (!this.branch) {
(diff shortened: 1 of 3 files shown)


src/project/git/GitCommandGitProject.ts
```diff
@@ -292,11 +301,14 @@ export class GitCommandGitProject extends NodeFsLocalProject implements GitProje
...
-               directoryManager: DirectoryManager,
-               secondTry: boolean = false): Promise<GitProject> {
+function clone(
+    credentials: ProjectOperationCredentials,
+    id: RemoteRepoRef,
+    opts: CloneOptions,
+    directoryManager: DirectoryManager,
+    secondTry: boolean = false,
+): Promise<GitProject> {
+
```
cdupuis: Why that change? Just looks like formatting. Maybe we need some agreement how we want that to be formatted.

cdupuis: Both good reasons. Thanks.

PR #297, opened 2018-05-27
message for a common error

jessitron:
this is leftover from the offsite, looks like it might still be useful


jessitron:
moved to https://github.com/atomisthq/automation-api/issues/437

PR #307, opened 2018-06-19
Pass visibility through to repo creation

jessitron:
it was getting lost.


jessitron:
mitigations that could have prevented this:
- validating that we never use `this` outside of a class (no-invalid-this in tslint?)
- validating that we type all parameters
- tests

PR #311, opened 2018-06-29
Add command to install kube utilities

### LINKED ISSUE #310: Have Atomist CLI install Kubernetes utilities
Provide an `atomist kube` command in the CLI that installs all Atomist Kubernetes utilities in the currently configured Kubernetes cluster, i.e., the cluster `kubectl` uses by default.

ddgenome:
Add `atomist kube` command that deploys k8vent and k8-automation to
the Kubernetes cluster using `kubectl`.

Closes #310

### DIFF
--- a/src/cli/kube.ts
+++ b/src/cli/kube.ts
@@ -1,0 +1,94 @@
+/*
+ * Copyright © 2018 Atomist, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


src/cli/kube.ts
```diff
@@ -0,0 +1,94 @@
...
+        return Promise.resolve(1);
+    }
+    const teamIds = resolveTeamIds(userConfig);
+    if (!teamIds || teamIds.length < 1) {
+        console.error(`No Atomist workspace/team IDs set, try running 'atomist config' first`);
+        return Promise.resolve(1);
+    }
+
+    const k8hookBase = `${webhookBaseUrl()}/atomist/kube/teams`;
+    const webhooks = `${k8hookBase}/` + teamIds.join(`,${k8hookBase}/`);
```
cdupuis: That's an interesting way for writing that. I think I would have done:

`teamsIds.map(id => ´${k8hookBase}/${id}´).join(",");`

ddgenome: That would be better. I had this solution in my head because I used it for Helm, whose templating system does not have map.

PR #321, opened 2018-07-23
HTTP client and factory abstraction

cdupuis:
This provides new `HttpClient` and `HttpClientFactory` interfaces. The latter can get registered on `configuration.http.client.factory` and overwrite the default of `AxiosHttpClientFactory`. 

I provided a `curl`-based implementation.

### DIFF
--- a/src/spi/http/curlHttpClient.ts
+++ b/src/spi/http/curlHttpClient.ts
@@ -1,0 +1,108 @@
+import * as _ from "lodash";
+import * as os from "os";
+import { logger } from "../..";
+import { doWithRetry } from "../../util/retry";
+import {
+    spawnAndWatch,
+    SuccessIsReturn0ErrorFinder,
+    WritableLog,
+} from "../../util/spawned";
+import {
+    DefaultHttpClientOptions,
+    HttpClient,
+    HttpClientFactory,
+    HttpClientOptions,
+    HttpResponse,
+} from "./httpClient";
+
+/**
+ * Curl based HttpClient implementation.
+ */
+export class CurlHttpClient implements HttpClient {
+
+    public exchange<T>(url: string,
+                       options: HttpClientOptions = {}): Promise<HttpResponse<T>> {
+
+        const optionsToUse: HttpClientOptions = {
+            ...DefaultHttpClientOptions,
+            ...options,
+        };
+
+        // Prepare headers
+        const headers = [];
+        _.forEach(optionsToUse.headers, (v, k) => {
+            headers.push(`${k}: ${v}`);
+        });
+
+        // Prepare the provided raw curl options
+        const rawOptions = [];
+        _.forEach(optionsToUse.options, (v, k) => {
+            if (k.length === 1) {
+                rawOptions.push([ `-${k}`, v ]);
+            } else {
+                rawOptions.push([ `--${k}`, v ]);
+            }
+        });
+
+        const request = () => {
+
+            let log = "";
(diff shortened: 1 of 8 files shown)


src/spi/http/curlHttpClient.ts
```diff
@@ -0,0 +1,108 @@
...
+        const headers = [];
+        _.forEach(optionsToUse.headers, (v, k) => {
+            headers.push(`${k}: ${v}`);
+        });
+
+        // Prepare the provided raw curl options
+        const rawOptions = [];
+        _.forEach(optionsToUse.options, (v, k) => {
+            if (k.length === 1) {
+                headers.push([ `-${k}`, v ]);
```
ddgenome: Shouldn't `headers` be `rawOptions`?

src/spi/http/curlHttpClient.ts
```diff
@@ -0,0 +1,108 @@
...
+            } else {
+                headers.push([ `--${k}`, v ]);
```
ddgenome: Same

src/spi/http/curlHttpClient.ts
```diff
@@ -0,0 +1,108 @@
...
+                       options: HttpClientOptions = {}): Promise<HttpResponse<T>> {
+
+        const optionsToUse: HttpClientOptions = {
+            ...DefaultHttpClientOptions,
+            ...options,
+        };
+
+        // Prepare headers
+        const headers = [];
+        _.forEach(optionsToUse.headers, (v, k) => {
```
ddgenome: Is there a reason you use `_.forEach`  and `push` instead of `_.map`?

test/spi/http/axiosHttpClientTest.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+import * as http from "http";
+import "mocha";
+import * as os from "os";
+import * as assert from "power-assert";
+import { AxiosHttpClientFactory } from "../../../src/spi/http/axiosHttpClient";
+import { HttpMethod } from "../../../src/spi/http/httpClient";
+
+const hostname = os.hostname();
```
ddgenome: I don't think this is reliable to bind against.

test/spi/http/axiosHttpClientTest.ts
```diff
@@ -0,0 +1,87 @@
...
+
+describe("axiosHttpClient", () => {
+
+    let server;
```
ddgenome: Would it be better to just have a server and close it in each test?

src/spi/http/curlHttpClient.ts
cdupuis: true.

src/spi/http/curlHttpClient.ts
cdupuis: true again.

src/spi/http/curlHttpClient.ts
cdupuis: `map` only works on arrays not on object literals.

test/spi/http/axiosHttpClientTest.ts
cdupuis: I tried a couple of different things and all work locally on two different boxes but not on the build pods.

src/spi/http/curlHttpClient.ts
ddgenome: `_.map` should work on any collection https://lodash.com/docs/#map

PR #323, opened 2018-07-24
Work around 'Type ... is not assignable to type Configuration'

jessitron:
When Configuration contains a class with private/protected fields, at any level,
then a Configuration that resolves to one installation of automation-client
is _never_ the same type that resolves to a Configuration that resolves o
another installation, even if they're the same version, if they're in different
spots on the filesystem. TypeScript protects us. :-/
So don't do it. It's better to never require a class, but only an interface.
Right now I want to see if this will get us around the problem.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -224,7 +224,7 @@
     },
     "@types/graphql": {
       "version": "0.12.7",
-      "resolved": "https://registry.npmjs.org/@types/graphql/-/graphql-0.12.7.tgz",
+      "resolved": "http://registry.npmjs.org/@types/graphql/-/graphql-0.12.7.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -224,7 +224,7 @@
     },
     "@types/graphql": {
       "version": "0.12.7",
-      "resolved": "https://registry.npmjs.org/@types/graphql/-/graphql-0.12.7.tgz",
+      "resolved": "http://registry.npmjs.org/@types/graphql/-/graphql-0.12.7.tgz",
```
cdupuis: I really don't understand why NPM is doing this all the time.

package-lock.json
```diff
@@ -2036,7 +2036,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "http://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
+      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
```
cdupuis: And this is the opposite.

PR #325, opened 2018-07-26
Replace run.ts with start.command.ts

ddgenome:
The current run.ts does not work with automation clients that rely on
scanning for command handlers.  Rather than fix that script, which
resides in the now deprecated cli subdirectory, I created
start.command.ts for two reasons.  First, it is easier to determine if
an automation client has a working script to run command handlers, it
is just a check for a file rather than having to parse it.  Second,
the new name parallels the start.client.ts script.

### DIFF
--- a/src/start.command.ts
+++ b/src/start.command.ts
@@ -1,0 +1,112 @@
+#!/usr/bin/env node
+/*
+ * Copyright © 2018 Atomist, Inc.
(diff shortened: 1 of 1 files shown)


jrday:

* 2 successful checks

[atomist:generated] [atomist:merge-on-approve]

PR #327, opened 2018-07-28
Do not send circular JSON back from event handlers

jessitron:
This was failing with TypeError in the response to /events.
It resulted in a socket hang up.

You may have a better way to defend against this.

### DIFF
--- a/src/internal/transport/AbstractRequestProcessor.ts
+++ b/src/internal/transport/AbstractRequestProcessor.ts
@@ -244,12 +244,20 @@ export abstract class AbstractRequestProcessor implements RequestProcessor {
                           callback: (results: Promise<HandlerResult[]>) => void) {
 
         const finalize = (results: HandlerResult[]) => {
-            this.sendEventStatus(results.some(r => r.code !== 0) ? false : true, ef, event, ctx)
+            let noncircularResults = results;
+            try {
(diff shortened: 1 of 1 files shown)


src/internal/transport/AbstractRequestProcessor.ts
```diff
@@ -244,14 +244,23 @@ export abstract class AbstractRequestProcessor implements RequestProcessor {
...
 
         const finalize = (results: HandlerResult[]) => {
-            this.sendEventStatus(results.some(r => r.code !== 0) ? false : true, ef, event, ctx)
+            let noncircularResults = results;
+            try {
+                JSON.stringify(noncircularResults);
+            } catch (err) {
+                logger.error("Circular object returned from event handler: %s", stringify(results));
+                noncircularResults = results.map(r => ({ code: r.code, message: stringify(r.message) }));
+                logger.error("Substituting for circular object: %j", noncircularResults);
```
ddgenome: Is logging statement needed since we log it either way below?

jessitron: I wanted to be explicit about the substitution that was made.

ddgenome: Doesn't the first log statement make it clear that a substitution will be made?

PR #329, opened 2018-07-31
Initial support for apiKeys

### LINKED ISSUE #308: Add support for API keys for registration using Authorization/Bearer

cdupuis:
This introduces two new configuration keys `apiKey` and `workspaceIds` to replace `token` and `teamIds`.

If an `apiKey` is specified, `token` will not be used for registration. If no `workspaceIds` are specified, `teamIds` will be copied into `workspaceIds` and used for registration. 

Also deprecated `teamId` on `HandlerContext` and a bunch of methods in `configuration.ts`.

Lastly I removed the CLI from this new version of the client.

fixes #308


src/configuration.ts
```diff
@@ -602,6 +616,9 @@ export function loadAtomistConfig(): AutomationServerOptions {
  * ATOMIST_TEAM environment variable, which takes precedence over the
  * configuration "teamdIds", which takes precedence over cfg.teamIds,
  * which may be undefined, null, or an empty array.
+ *
+ * @deprecated The usage of teamIds is deprecated and will be removed
```
ddgenome: Should this refer to `resolveTeamIds` instead of just `teamIds`?

src/configuration.ts
```diff
@@ -614,6 +631,24 @@ export function resolveTeamIds(cfg: Configuration): string[] {
...
+ * Examine environment, config, and cfg for Atomist workspace IDs.  The
+ * ATOMIST_WORKSPACES environment variable takes precedence over the
+ * ATOMIST_WORKSPACE environment variable, which takes precedence over the
+ * configuration "workspaceIds", which takes precedence over cfg.workspaceId,
+ * which may be undefined, null, or an empty array.
+ */
+export function resolveWorkspaceIds(cfg: Configuration): string[] {
+    if (process.env.ATOMIST_WORKSPACES) {
+        cfg.workspaceIds = process.env.ATOMIST_WORKSPACES.split(",");
+    } else if (process.env.ATOMIST_WORKSPACE) {
```
ddgenome: Can we take this opportunity to remove the singular version? I think it is more confusing to support two environment variables than it is to explain that the one we do support can accept a comma-delimited list of IDs.

src/configuration.ts
```diff
@@ -649,6 +684,9 @@ export function resolveConfigurationValue(
  * ATOMIST_TOKEN environment variable takes precedence over the
  * GITHUB_TOKEN environment variable, which takes precedence over the
  * config value, which takes precedence over the passed in value.
+ *
+ * @deprecated The usage of tokens is deprecated and will be removed
```
ddgenome: resolveToken?

src/configuration.ts
```diff
@@ -801,9 +851,7 @@ function validateConfiguration(cfg: Configuration) {
  * cause any values provided by sources of lower precedence to be
  * ignored.  Arrays are replaced, not merged.  Typically the only
  * required values in the configuration for a successful registration
- * are the token and non-empty teamIds.  These can be provided via the
- * ATOMIST_TOKEN and ATOMIST_TEAMS environment variables,
- * respectively.
+ * are the apiKey or token and non-empty teamIds.
```
ddgenome: s/teamIds/workspaceIds/

src/internal/invoker/Payload.ts
```diff
@@ -5,9 +5,32 @@ import { Source } from "../transport/RequestProcessor";
...
-    teamId: string;
+    /**
+     * ID of the Atomist Workspace
+     * @deprecated use workspaceId instead
+     */
+    teamId?: string;
+
+    /**
+     * ID of the Atomist Workspace
+     */
```
ddgenome: Short TypeDoc like this can use the single line form: `/** ID of the Atomist workspace */`

src/internal/transport/AbstractRequestProcessor.ts
```diff
@@ -299,8 +301,8 @@ export abstract class AbstractRequestProcessor implements RequestProcessor {
                              ts: number = Date.now()) {
         return {
             correlationId: _.get(request, "correlation_id") || _.get(request, "extensions.correlation_id"),
-            teamId: _.get(request, "team.id") || _.get(request, "extensions.team_id"),
-            teamName: _.get(request, "team.name") || _.get(request, "extensions.team_name"),
+            workspaceId: _.get(request, "team.id") || _.get(request, "extensions.team_id"),
+            workspaceName: _.get(request, "team.name") || _.get(request, "extensions.team_name"),
```
ddgenome: Has the request object not been updated to use workspace?

test/util/git.ts
```diff
@@ -0,0 +1,38 @@
+import { SpawnOptions } from "child_process";
+import * as fs from "fs-extra";
+import "mocha";
+import * as path from "path";
+import * as assert from "power-assert";
+import { obtainGitInfo } from "../../src/internal/env/gitInfo";
+
```
ddgenome: Where did this come from? Why is the source and the test in this file under test?

cdupuis: I needed a way to create the git-info.json but can’t rely on cli ad that depends on client. I wanted something that doesn’t end up in our source tree.

src/internal/transport/AbstractRequestProcessor.ts
cdupuis: No

src/configuration.ts
cdupuis: Fixing

src/configuration.ts
cdupuis: Fixing

src/configuration.ts
cdupuis: Will do

src/configuration.ts
cdupuis: Fixing

PR #333, opened 2018-08-03
ExportMoreThingsTransform

jessitron:
ExportMoreThingsTransform

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,100 +16,430 @@
  */
 
 export {
-    AtomistBuild,
-    AtomistLinkImage,
-    AtomistWebhookType,
-    postAtomistWebhook,
-    webhookBaseUrl,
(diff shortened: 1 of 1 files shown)


jessitron:
@cd this is the result of an editor.

Would you like any changes to the formatting? I can change the code and regenerate it.

jessitron:
The duplicate identifiers here, I think, are worth fixing by hand.

jessitron:
There are also some that aren't valid, that are imported by atomisthq projects that use old versions of client. For instance: https://github.com/atomisthq/pull-request-semantic-differ/blob/db8411a22615337d995cef2002e4031501597093/src/ingest/PullRequestIngestor.ts

jessitron:
then there's stuff like Parameter, which we really define two of.

PR #335, opened 2018-08-08
Use name to lookup graphql by operation name

### LINKED ISSUE #334: Feedback on graphClient.executeQueryFromFile versus graphClient.query
```
ctx.graphClient.executeQueryFromFile<PodDeployments.PodDeployment, PodDeployments.Variables>(
        "./src/graphql/query/podDeployments.graphql",
        {
        });
```
versus

```
ctx.graphClient.query<PodDeployments.PodDeployment, PodDeployments.Variables>(
       {name: "podDeployments",
        options: {}}
        );
```

I think the latter is preferred, where you use the convention for how graphql query files are laid out in the underlying node module.  

There is still a potential difference between the name of the graphql file and and the name of the graphql query in that file.  For some reason, I was sort of expecting those graphql files to have been indexed by the  names of the queries they contain so it was only on my second run through that I tried using the query's filename.  

I also had an issue where I had a mutation graphql in the query directory and then had some difficult figuring out that I needed to use the `mutation` api, and move the file into the other structure.  The convention seems good but I think there are some important missing docs here.  The GraphClient interface mentions " Run query either provided as a string or as loaded from a file" but the QueryOptions doesn't have enough context to figure out this convention. 

The exmaples here:  https://docs.atomist.com/developer/graphql/ do not refer to the `query` api so also wnat to confirm that we should be using this one.

cdupuis:
fixes #334

### DIFF
--- a/src/internal/graph/graphQL.ts
+++ b/src/internal/graph/graphQL.ts
@@ -10,6 +10,7 @@ import {
 import gql from "graphql-tag";
 import { validate } from "graphql/validation";
 import * as p from "path";
+import { logger } from "../..";
 import {
     findLine,
     generateHash,
@@ -330,7 +331,7 @@ function locateAndLoadGraphql(
 ): string {
 
     let path = options.path;
-    let name = options.name;
+    const name = options.name;
     // Read subscription from file if given
     if (options.path) {
         if (!path.endsWith(".graphql")) {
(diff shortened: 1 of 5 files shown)


src/spi/graph/GraphClient.ts
```diff
@@ -63,28 +60,63 @@ export interface GraphClient {
      * @param {Q} variables the variables to be used
      * @param {any} options to be passed onto the underlying implementation
      * @returns {Promise<T>}
-     *
-     * DEPRECATED: use mutate() instead
+     * @deprecated use mutate() instead
      */
     executeMutation<T, Q>(mutation: string, variables?: Q, options?: any): Promise<any>;
```
ddgenome: I know you didn't change this, but shouldn't this return `Promise<T>`?

src/internal/graph/graphQL.ts
```diff
@@ -343,12 +344,30 @@ function locateAndLoadGraphql(
...
+                logger.warn(`WARNING: no ${subfolder} graphql operation found for name '${options.name}. ` +
+                    `Falling back to file name lookup. Support for file name lookup will be removed in a future release.'`);
+                if (!options.name.endsWith(".graphql")) {
+                    path = p.join(queryDir, `${options.name}.graphql`);
+                } else {
+                    path = p.join(queryDir, options.name);
+                }
+                // End of remove for next major release
+                // throw new Error(`No matching ${subfolder} graphql operation found for name '${options.name}'`);
+            } else if (queries.length > 1) {
```
ddgenome: `else` would probably suffice here.

src/spi/graph/GraphClient.ts
```diff
@@ -63,28 +60,63 @@ export interface GraphClient {
...
+
+    /** Optional GraphQL query string */
     query?: string;
+    /**
+     * Optional absolute or relative path to the '.graphql' query file.
+     * Relative paths will get resolved against the path of the calling script.
+     */
     path?: string;
+    /**
+     * Optional name of the query operation in a '.graphql' file located in 'graphql/query'.
```
ddgenome: One must specify either `query` or `name`, correct?

src/spi/graph/GraphClient.ts
```diff
@@ -63,28 +60,63 @@ export interface GraphClient {
...
+
+    /** Optional GraphQL mutation string */
     mutation?: string;
+    /**
+     * Optional absolute or relative path to the '.graphql' query file.
+     * Relative paths will get resolved against the path of the calling script.
+     */
     path?: string;
+    /**
+     * Optional name of the mutation operation in a '.graphql' file located in 'graphql/mutation'.
```
ddgenome: One of `mutation` or `query` must be provided, correct?

cdupuis: Add a interface-level comment to reflect that.

src/spi/graph/GraphClient.ts
cdupuis: Add a interface-level comment to reflect that.

PR #347, opened 2018-08-28
If no event handlers are found, return 404 not 500

jessitron:
Then I am trying to make sdm-local treat 404 as OK from this endpoint; it's OK if there are no handlers


jessitron:
This is one of those PRs where I won't be surprised if you refactor it. I did minimal changes because I don't know what impact it would have for the function to return an empty array of results instead of throwing (I think that makes more sense)

PR #350, opened 2018-08-31
Remove all deprecations

cdupuis:
Removed all the deprecations except a few that were being used in a lot of places.

Sign off from everybody on the reviewers list required for this PR to get merged.


test/api/ApolloGraphClientTest.ts
```diff
@@ -16,71 +15,6 @@ import {
 
 describe("ApolloGraphClient", () => {
 
-    describe("deprecated api", () => {
```
ddgenome: I didn't notice this before, but this was smart to do.

cdupuis: I don't think I did that. Not that clear. ;-)

src/configuration.ts
```diff
@@ -670,7 +622,7 @@ export function resolveTeamIds(cfg: Configuration): string[] {
  * @return the resolved workspace IDs
  */
 export function resolveWorkspaceIds(cfg: Configuration): string[] {
-    const teamIds = resolveTeamIds(cfg);
+    const teamIds = [];
     if (process.env.ATOMIST_WORKSPACES) {
         cfg.workspaceIds = process.env.ATOMIST_WORKSPACES.split(",");
     } else if (config("workspaceIds")) {
```
jessitron: there's an if statement here that can go away, that checks the length on teamIds[]

src/operations/common/allReposInTeamRepoFinder.ts
```diff
@@ -15,9 +15,6 @@ const PageSize = 100;
 /**
  * Use a GraphQL query to find all repos for the current team,
  * or look locally if appropriate, in current working directory
- *
- * DEPRECATED: there's a better one in @atomist/sdm
- *
```
jessitron: wait why is this deprecation removed? This one is actually dangerous, because it works off the current directory, which does not fit the SDM's way of working, and people could use this accidentally and get very confused. How about we remove this one instead?

cdupuis: It is being used in a lot of places. This for finding repos in graphql, not on the local file system.

src/configuration.ts
cdupuis: Good catch. Will remove.

cdupuis:
Also `generatorToCommand` needs to come back. It is being used in SDM and we need it.

src/operations/common/allReposInTeamRepoFinder.ts
jessitron: I pushed a commit. Tests don't run locally for me on this project

PR #356, opened 2018-09-04
Add hasFile method to Project

johnsonr:
More convenient for users


ddgenome:
Thinking a bit more about this, should this method have been called `fileExists` to be parallel with the other methods, including `fileExistsSync`?

cdupuis:
We went with hasFile because we already have a similar push test in the SDM.

PR #360, opened 2018-09-06
Add Project.hasDirectory, fix file tests & leaks

ddgenome:
Add hasDirectory to Project interface and implementations for
NodeFsLocalProject and InMemoryProject.  Fix file tests to only return
true/Files for regular files.  Use path.join() in NodeFsLocalProject.
Improve Project TypeDoc.  Add many tests.

Many tests left temporary directories around, many from
GitCommandGitProject.cloned().  Hunt these down and ensure the
directories are cleaned up.  Convert tests to async as I went.  Move
helper functions into their own file so tests do not run multiple
times or when not intended.

The test directory clean up turned up a few actual leaks when using
`generate`.  It is not possible to delete the generated projects
directory using release().  When generate failed, there was _no_ way
to delete the directory it created.  I fixed the latter but the former
will require some more digging.  The simple fix resulted in a stack
overflow.


ddgenome:
I went with `hasDirectory` instead of `isDirectory` because the latter is typically called on a `stat` object and the former has a parallel to the already existing `hasFile`.

PR #370, opened 2018-09-12
Update to Winston 3 and make logging configuration explicit

cdupuis:


src/bin/gql-gen.ts
```diff
@@ -53,6 +51,7 @@ function libDir(cwd: string): string {
  * Generate TypeScript typings for GraphQL schema entities.
  */
 async function main(): Promise<void> {
+    configureLogging(NoLoggingConfiguration);
```
ddgenome: Why wouldn't `NoLoggingConfiguration` just be the default and then we can omit all this nonsense?

src/index.ts
```diff
@@ -88,10 +89,7 @@ export {
     CommandInvocation,
 } from "./internal/invoker/Payload";
 
-export {
-    logger,
-    LoggingConfig,
-} from "./internal/util/logger";
+export * from "./internal/util/logger";
```
ddgenome: Seems odd to export stuff from under internal. Perhaps this does not belong under internal?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+         * Set format for file logging
+         */
+        format?: LoggingFormat,
+    };
+}
+
+/**
+ * Logging configuration to simply pass through log message to the console
+ */
+export const NoLoggingConfiguration: LoggingConfiguration = {
```
ddgenome: `NoLoggingConfiguration` seems like a misnomer to me. When I first saw it above I thought it meant all log messages would be eaten, i.e., no log statements would be output. Perhaps `PlainLoggingConfiguration` would be better?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
 
-export function setLogLevel(level: string) {
-    winstonLogger.transports.console.level = level;
+/**
+ * Constants for the logging format
+ */
+export enum LoggingFormat {
+    /**
+     * Print only the message provided to the logger methods
+     */
```
ddgenome: If you want to condense this a bit, you can enter the one-line TypeDoc like this:

```typescript
/** Print only the message ... */
```

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+                `${pj.name.replace(/^.*\//, "")}-local.log`);
+        }
+
+        const path = p.resolve(filename);
+        fs.mkdirsSync(p.dirname(path));
+
+        const ft = new winston.transports.File({
             filename: p.basename(path),
             dirname: p.dirname(path),
+            level: config.file.level || config.console.level,
```
ddgenome: Perhaps this should be set to the greater of the two values rather than the first truthy value?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+        winstonLogger.silent = false;
+    }
+
+    // Set up file logging
+    if (config.file.enabled) {
+
+        let filename = config.file.filename;
+
+        if (!filename) {
+            const appDir = __dirname.split(p.join("node_modules", "@atomist", "automation-client"))[0];
```
ddgenome: Remind me, we're doing this because read-pkg-up doesn't work when things are globally installed?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+            let pj: { name: string; };
+            try {
+                pj = require(p.join(appDir, "package.json"));
+            } catch (e) {
+                pj = { name: "atm-client" };
+            }
+            filename = p.join(
+                os.homedir(),
```
ddgenome: I did not know that existed.

Note to self: use that when locating the user configuration file.

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+                ".atomist",
+                "log",
+                `${pj.name.replace(/^.*\//, "")}-local.log`);
+        }
+
+        const path = p.resolve(filename);
+        fs.mkdirsSync(p.dirname(path));
```
ddgenome: Yes, please.

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+ * Configure the logging sub-system based on the provided Configuration object
+ * @param configuration
+ */
+export function clientLoggingConfiguration(configuration: Configuration): LoggingConfiguration {
+    const lc: LoggingConfiguration = {
+        ...DefaultClientLoggingConfiguration,
+    };
+    if (configuration.logging) {
+        if (configuration.logging.level) {
+            lc.console.enabled = true,
```
ddgenome: Why is there a comma at the end of this line?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+                lc.console.level = configuration.logging.level;
+            lc.console.format = LoggingFormat.Full;
+        }
+        if (configuration.logging.file) {
+            if (configuration.logging.file.enabled === true) {
```
ddgenome: Seems like you could combine these two if statements into one.

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+                let filename = p.join(".", "log", `${configuration.name.replace(/^.*\//, "")}.log`);
```
ddgenome: I'm pretty sure starting the path with "." is not necessary. It also seems you could use a const and a ternary operator here.

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
-export function formatter(options: any): string {
-    if (LoggingConfig.format === "cli") {
-        return options.message;
+/**
+ * Configure the logging sub-system based on the provided Configuration object
+ * @param configuration
+ */
+export function clientLoggingConfiguration(configuration: Configuration): LoggingConfiguration {
+    const lc: LoggingConfiguration = {
+        ...DefaultClientLoggingConfiguration,
```
ddgenome: Does this protect modification of nested properties or does it result in a shallow copy?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+                    format: LoggingFormat.Full,
+                };
+            }
+        }
     }
+    return lc;
+}
 
+const clientFormat = info => {
+    const c = winston.format.colorize();
```
ddgenome: So calling the function colorize returns an object that has a function colorize?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
     const executionContext = context.get();
     let ctx: string;
     if (cluster.isMaster) {
-        ctx = options.colorize ? winston.config.colorize(options.level, "m")
```
ddgenome: Is this conditional handled some other way now?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
+                };
+            }
+        }
     }
+    return lc;
+}
 
+const clientFormat = info => {
+    const c = winston.format.colorize();
     const executionContext = context.get();
```
ddgenome: I did not know you could get the context like this.

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
-        + " [" + level + "] " + (options.message ? options.message : "") +
-        (options.meta && Object.keys(options.meta).length ?
-            (options.message ? ": " : "") + stringify(options.meta) : "");
+    const formatted = info.timestamp + (ctx ? " [" + ctx + "]" : "")
+        + " [" + level + "] " + (info.message ? info.message : "");
 
-    if (options.colorize) {
-        return formatted;
-    } else {
-        return stripAnsi(formatted);
```
ddgenome: Maybe this is a way to handle the colorize conditional?

src/internal/util/logger.ts
```diff
@@ -1,119 +1,286 @@
...
-            showLevel: true,
-            align: true,
-            stderrLevels: ["error"],
-            formatter,
-            // handleExceptions: true,
-            // humanReadableUnhandledException: true,
-        }),
-    ],
-});
+function unRedirectConsoleLogging() {
```
ddgenome: I like this.

test/api/apiUtils.ts
```diff
@@ -16,26 +16,17 @@
...
-    GitHubDotComBase,
-    GitHubRepoRef,
-} from "../../src/operations/common/GitHubRepoRef";
+import { GitHubDotComBase } from "../../src/operations/common/GitHubRepoRef";
 import { ProjectOperationCredentials } from "../../src/operations/common/ProjectOperationCredentials";
 import { RemoteRepoRef } from "../../src/operations/common/RepoId";
 import { LocalProject } from "../../src/project/local/LocalProject";
 import { TestRepositoryVisibility } from "../credentials";
 
-LoggingConfig.format = "cli";
```
ddgenome: So is the mocha output all clean now?

src/bin/gql-gen.ts
cdupuis: Do you think the straight to console.log is a good default? I think it probably is.

src/index.ts
cdupuis: That's right. I'll move it to `util`.

src/internal/util/logger.ts
cdupuis: Right, I'll change it.

src/internal/util/logger.ts
cdupuis: Yeah, I'm not that sure any longer. Also I created  that other helper for this. I'll revisit.

src/internal/util/logger.ts
cdupuis: Good question.

src/internal/util/logger.ts
cdupuis: Yes.

cdupuis: Winston now ships types so that is call compile-time checked.

src/internal/util/logger.ts
cdupuis: Yeah, internally of winston it doesn't colorise if is turned off.

test/api/apiUtils.ts
cdupuis: Yes, clean as it can be.

src/internal/util/logger.ts
cdupuis: No, the intent is to take the level of the file configuration if it exists, if not fall back to console's level.

ddgenome:
Thinking a bit more about the defaults. Perhaps in command-line tools we should use the various `console` methods, avoiding `logger`. Then we could keep the default `logger` behavior what you want, no output. Then we would only override the `console` methods when logging is configured. What do you think?

cdupuis:
I don't have strong sense of what is right or wrong here. For some of our scripts it is not always clear to me if that is just being used as a cli command or also from code. sdm-local especially is full of that.

cdupuis:
@jessitron I’m going to merge this now to be able to fix all the downstream consumers. But please review the changes in logger.ts.

PR #372, opened 2018-09-12
Remove CLS and use asyncHooks

cdupuis:


src/internal/util/cls.ts
```diff
@@ -1,27 +1,17 @@
-import {
-    createNamespace,
-    getNamespace,
-    Namespace,
-} from "continuation-local-storage";
+import * as asyncHooks from "async_hooks";
 
-// create the local storage namespace
-createNamespace("automation-client");
+const namespaces = {};
```
ddgenome: Is this of type `Namespace`?

src/internal/util/cls.ts
```diff
@@ -1,27 +1,17 @@
...
 
-export function init(): Namespace {
-    return getNamespace("automation-client");
+export function create() {
```
ddgenome: Should this return `: Namespace`?

src/internal/util/cls.ts
```diff
@@ -36,3 +26,62 @@ export interface AutomationContext {
     ts: number;
 
 }
+
+class Namespace {
+
+    constructor(private readonly context = {}) {
+    }
+
+    public run(fn) {
```
ddgenome: Is there a reason none of these functions have their return type specified.

cdupuis:
Sorry, this code is highly experimental. Didn’t intent to raise a PR yet. Needs a lot more testing to see if it actually works.

PR #382, opened 2018-09-18
Export cross-spawn as spawn

ddgenome:
So we can have a common spawn and exec to run from SDMs/clients.


lib/util/spawn.ts
```diff
@@ -25,6 +25,8 @@ import { sprintf } from "sprintf-js";
 import * as strip_ansi from "strip-ansi";
 import { logger } from "./logger";
 
+export { spawn };
```
lievendoclo: Why not export this in the index?

PR #398, opened 2018-10-30
Update the API Doc to reflect configuration load from index.ts

jessitron:


jessitron:
that build failure (a test timeout) has nothing to do with my changes

PR #409, opened 2018-11-29
Extract creation of HTTP and WS clients for registration to proper factories

cdupuis:
With this change, HTTP and WS client can be configured correctly to work
in all sorts of network situations, like tunnelling proxies.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -42,6 +42,13 @@ export {
     automationClientInstance,
 } from "./lib/globals";
 import * as GraphQL from "./lib/graph/graphQL";
+import * as validationPatterns from "./lib/operations/common/params/validationPatterns";
+import * as editModes from "./lib/operations/edit/editModes";
+import * as parseUtils from "./lib/project/util/parseUtils";
+import * as projectUtils from "./lib/project/util/projectUtils";
+import * as secured from "./lib/secured";
+import * as astUtils from "./lib/tree/ast/astUtils";
+
 export { GraphQL };
 export {
     EventFired,
@@ -100,7 +107,6 @@ export {
 export {
     RemoteLocator,
 } from "./lib/operations/common/params/RemoteLocator";
-import * as validationPatterns from "./lib/operations/common/params/validationPatterns";
 export { validationPatterns };
 export {
     ProjectOperationCredentials,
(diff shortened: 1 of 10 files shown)


index.ts
```diff
@@ -42,6 +42,13 @@ export {
     automationClientInstance,
 } from "./lib/globals";
 import * as GraphQL from "./lib/graph/graphQL";
+import * as validationPatterns from "./lib/operations/common/params/validationPatterns";
+import * as editModes from "./lib/operations/edit/editModes";
+import * as parseUtils from "./lib/project/util/parseUtils";
+import * as projectUtils from "./lib/project/util/projectUtils";
+import * as secured from "./lib/secured";
+import * as astUtils from "./lib/tree/ast/astUtils";
```
ddgenome: Why move all these? And if you are going to break the alphabetical order, why not put them all at the top or bottom?

lib/internal/transport/websocket/WebSocketClient.ts
```diff
@@ -218,15 +211,21 @@ function register(registrationCallback: () => any,
...
-            timeout: configuration.ws.timeout || 10000,
-        };
 
-        return axios.post(configuration.endpoints.api, registrationPayload, configureProxy(config))
+        return client.exchange<RegistrationConfirmation>(configuration.endpoints.api, {
+                body: registrationPayload,
+                method: HttpMethod.Post,
+                headers: { Authorization: authorization },
+                options: {
+                    timeout: configuration.ws.timeout || 10000,
```
ddgenome: Shouldn't this default be provided by the configuration defaults?

lib/util/retry.ts
```diff
@@ -24,16 +29,18 @@ export const DefaultRetryOptions: WrapOptions = {
...
 export function doWithRetry<R>(what: () => Promise<R>,
                                description: string,
-                               opts: WrapOptions = {}): Promise<R> {
+                               opts: RetryOptions = {}): Promise<R> {
     const retryOptions: WrapOptions = {
-        ...DefaultRetryOptions,
+        ...DefaultRetryOptions as WrapOptions,
         ...opts,
     };
     logger.log("silly", `${description} with retry options '%j'`, retryOptions);
```
ddgenome: Should this be inside a `opts.log` conditional?

PR #410, opened 2018-12-06
Add `GraphClientFactory`

cdupuis:
This allows to modify the creation of `GraphClient` instances through a factory. 

The default implementation `ApolloGraphClientFactory` allows to plug in strategies to modify the internally used fetch instance.

### DIFF
--- a/lib/internal/transport/RequestProcessor.ts
+++ b/lib/internal/transport/RequestProcessor.ts
@@ -31,6 +31,15 @@ export function isEventIncoming(event: any): event is EventIncoming {
     return !!event.data;
 }
 
+export function workspaceId(event: CommandIncoming | EventIncoming): string | undefined {
+    if (isCommandIncoming(event)) {
+        return event.team.id;
+    } else if (isEventIncoming(event)) {
+        return event.extensions.team_id;
+    }
+    return undefined;
+}
+
 export interface EventIncoming {
 
     data: any;
--- a/lib/internal/transport/express/ExpressRequestProcessor.ts
+++ b/lib/internal/transport/express/ExpressRequestProcessor.ts
@@ -9,6 +9,7 @@ import {
 import { AutomationEventListener } from "../../../server/AutomationEventListener";
 import { AutomationServer } from "../../../server/AutomationServer";
 import { GraphClient } from "../../../spi/graph/GraphClient";
+import { GraphClientFactory } from "../../../spi/graph/GraphClientFactory";
 import {
     Destination,
     MessageClient,
@@ -20,6 +21,7 @@ import {
     EventIncoming,
     isCommandIncoming,
     isEventIncoming,
+    workspaceId,
 } from "../RequestProcessor";
 import {
     WebSocketCommandMessageClient,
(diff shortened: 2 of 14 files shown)


lib/internal/transport/express/ExpressRequestProcessor.ts
```diff
@@ -47,9 +48,18 @@ export class ExpressRequestProcessor extends AbstractRequestProcessor {
 
     protected createGraphClient(event: EventIncoming | CommandIncoming,
                                 context: AutomationContextAware): GraphClient {
+        let workspaceId;
+        if (isCommandIncoming(event)) {
+            workspaceId = event.team.id;
+        } else if (isEventIncoming(event)) {
+            workspaceId = event.extensions.team_id;
+        }
```
ddgenome: This logic appears twice in this PR. Should it be centralized and explained?

lib/internal/transport/websocket/DefaultWebSocketRequestProcessor.ts
```diff
@@ -79,7 +78,16 @@ export class DefaultWebSocketRequestProcessor extends AbstractRequestProcessor
     }
 
     protected createGraphClient(event: CommandIncoming | EventIncoming): GraphClient {
-        return this.graphClients.createGraphClient(event);
+        let workspaceId;
+        if (isCommandIncoming(event)) {
+            workspaceId = event.team.id;
+        } else if (isEventIncoming(event)) {
+            workspaceId = event.extensions.team_id;
+        }
```
ddgenome: Three times.

lib/internal/transport/websocket/DefaultWebSocketRequestProcessor.ts
```diff
@@ -79,7 +78,16 @@ export class DefaultWebSocketRequestProcessor extends AbstractRequestProcessor
...
+        return this.graphClients.create(
+            workspaceId,
+            this.configuration,
+            () => `Bearer ${this.registration.jwt}`);
```
ddgenome: And this twice.

cdupuis:
Good points. Thanks. Fixes coming.

lib/internal/transport/express/ExpressRequestProcessor.ts
ddgenome: MIssed one.

lib/internal/transport/RequestProcessor.ts
```diff
@@ -31,6 +31,14 @@ export function isEventIncoming(event: any): event is EventIncoming {
     return !!event.data;
 }
 
+export function workspaceId(event: CommandIncoming | EventIncoming): string {
+    if (isCommandIncoming(event)) {
+        return event.team.id;
+    } else if (isEventIncoming(event)) {
+        return event.extensions.team_id;
+    }
```
ddgenome: `else`?

cdupuis: I wondered about that. If I do that, it means that anything that you would force into the function via `as any` would end up in that branch and probably fail with `can not read team_id of undefined`.

ddgenome: As it stands, the function can return nothing, which will likely screw up downstream code. It should probably explicitly return `undefined` and caller have to deal with that.

PR #417, opened 2018-12-07
Introduce promise pool to limit number of concurrent executions

cdupuis:


lib/util/pool.ts
```diff
@@ -0,0 +1,34 @@
...
+function concurrentDefault(): number {
+    return configurationValue<number>("pool.concurrent", 5);
+}
+
+/**
+ * Execute all provided promises with a max concurrency
+ * @param promises all promises to execute
+ * @param concurrent the max number of concurrent promise executions
+ */
+export async function executeAll<T>(promises: Array<Promise<T>>, concurrent: number = concurrentDefault()): Promise<T[]> {
```
ddgenome: I did not know you could use a call a function to provide a default value. I guess because that is all actually managed at run time, not compile time.

lib/util/pool.ts
```diff
@@ -0,0 +1,34 @@
...
+            return null;
+        }
+    };
+
+    const PromisePool = require("es6-promise-pool");
+    const pool = new PromisePool(producer, concurrent);
+
+    const results: T[] = [];
+    pool.addEventListener("fulfilled", (event: any) => {
+        results.push(event.data.result);
```
ddgenome: Couldn't this approach return the results in a different order than the original array of promises? You could pretty easily test this.

lib/util/pool.ts
```diff
@@ -0,0 +1,34 @@
...
+    });
+    await pool.start();
```
ddgenome: Although it would probably just be easier scrap the listener and capture the return value here and return that. Or just `return pool.start();`

lib/util/pool.ts
cdupuis: I added a different strategy to fix this. Good point.

lib/util/pool.ts
cdupuis: `pool.start()` doesn't return the results. I thought that too but it isn't the case.

ddgenome: That's terrible.

lib/util/pool.ts
```diff
@@ -0,0 +1,41 @@
...
+    };
+
+    const PromisePool = require("es6-promise-pool");
+    const pool = new PromisePool(producer, concurrent);
+
+    pool.addEventListener("fulfilled", (event: any) => {
+        results[results.indexOf(event.data.promise)] = event.data.result;
+    });
+    pool.addEventListener("rejected", (event: any) => {
+        results[results.indexOf(event.data.promise)] = event.data.error;
```
ddgenome: `Promise.all` rejects immediately on any rejected promise. Are you intentionally choosing different behavior?

cdupuis: I think I’m our usecaes you want to actually exit early. Keep going and collect. Imho that makes more sense. 

Maybe I should add the `Promise<Array<T | Error>>` as return value. 

What you recommend?

cdupuis: Turns out the `es6-promise-pool` rejects immediately on any rejected promise. So we are not changing the semantics here.

Actually I don't like it that much but that is for another day.

PR #430, opened 2018-12-20
Performance optimization in AST parsing

johnsonr:
Optimize, determining whether a file needs to be parsed based on literal value tests using string searches before invoking an actual `FileParser`.

When using expensive parsers such as ANTLR, this can increase performance by many times greatly reducing the number of files that need to be parsed.

Also add an optional method `FileParser` implementations can implement to provide further optimization.


johnsonr:
Superseded. All these changes will be part of next PR.

PR #434, opened 2018-12-20
Introduce `targets.branch` and fix `project.id` to point to correct project

cdupuis:
Introduce `targets.branch` to instruct a `CodeTransform` to work off of a branch.

Fix `project.id` as 1st parameter to a `CodeTransform` to always point to the correct repo; and not - in the case of project creation - to the local cloned seed repo.

Removed default to merge into `master` for `PullRequest` edit mode.

### DIFF
--- a/lib/internal/transport/websocket/WebSocketLifecycle.ts
+++ b/lib/internal/transport/websocket/WebSocketLifecycle.ts
@@ -70,9 +70,7 @@ export class WebSocketLifecycle {
             while (this.messages.length) {
                 queuedMessages.push(this.messages.pop());
             }
-            for (const queuedMessage of queuedMessages) {
-                await this.send(queuedMessage);
-            }
+            queuedMessages.forEach(this.send);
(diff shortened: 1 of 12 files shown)


lib/internal/transport/websocket/WebSocketLifecycle.ts
```diff
@@ -70,9 +70,7 @@ export class WebSocketLifecycle {
             while (this.messages.length) {
                 queuedMessages.push(this.messages.pop());
             }
-            for (const queuedMessage of queuedMessages) {
-                await this.send(queuedMessage);
-            }
+            queuedMessages.forEach(this.send);
```
ddgenome: Do you not want to `await` these?

```typescript
await Promise.all(queuedMessages.forEach(this.send));
```

cdupuis: No, `this.send` is void.

PR #446, opened 2019-01-09
Allow boolean and number parameters types in command requests

dansmithy:


cdupuis:
I'll rebase my changes onto your branch and push. Then you can review.

lib/internal/parameterPopulation.ts
```diff
@@ -57,22 +58,24 @@ function computeValue(parameter: { name: string, type?: ParameterType }, value:
             break;
         case "boolean":
             if (typeof value !== "boolean") {
-                value = value === "true" || value === "yes" || value === "1";
+                newValue = value === "true" || value === "yes" || value === "1";
             }
             break;
         case "number":
             if (typeof value === "string") {
                 value = parseInt(value, 10);
```
dansmithy: surely this should be `newValue =` ... does the below test really pass? (I can make this change if not)

cdupuis: Yeah, that was a rebase/merge error that I didn't catch.

PR #454, opened 2019-01-12
Allow path expressions to be backed by regular expressions with capture groups

johnsonr:
Similar to microgrammar based parsers, make it possible to match and destructure regular expression groups using path expressions. For example:

```typescript
const personParser = new RegexFileParser({
    rootName: "people",
    matchName: "person",
    regex: /([A-z][a-z]+):([\d]+)/,
    captureGroupNames: ["name", "age"],
});

const matches = await findMatches<Person>(p, personParser, AllFiles, "/people/person/name");
```

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -243,6 +243,12 @@ export {
     TypeScriptES6FileParser,
 } from "./lib/tree/ast/typescript/TypeScriptFileParser";
 export {
+    MicrogrammarBasedFileParser,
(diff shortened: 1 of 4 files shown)


jessitron:
What is the use case? (I can think of some but want yours)

johnsonr:
Doing consistent regexp replaces (or taking an action per match) is otherwise somewhat complicated. This way, it's easy and consistent with microgrammar and other AST usage.

jessitron:
Can I have a more specific example please so I can document it?

PR #474, opened 2019-02-15
Update all deps including graphql-code-generator

cdupuis:
This updates graphql-code-generator to latest version which needed a couple of changes:

* now using the api to generate because otherwise we would need to have `codegen.yml` file in each repo 👎 
* I had to patch the web packed handlebar template to get the old bahavior back that generates optional properties on the generated types. Without that change, we have many compile errors in sdm and sdm-core then further.

Should we move forward with this for now? The motivation for the upgrade was a security vulnerability
on one of the old transitive dependencies. WDYT?


bin/gql-gen.ts
```diff
@@ -40,53 +40,54 @@ async function main(): Promise<void> {
         const cwd = process.cwd();
         const lib = libDir(cwd);
 
+        // patch up codegen template to create optional properties
+        const codegenTemplate = path.join(cwd, "node_modules", "graphql-codegen-typescript-client", "dist", "index.js");
+        if (await fs.pathExists(codegenTemplate)) {
+            let contents = (await fs.readFile(codegenTemplate)).toString();
+            contents = contents.replace(
+                /{{ name }}: {{ convertedFieldType/i,
+                "{{ name }}?: {{ convertedFieldType");
```
ddgenome: If this regular expression is case insensitive, shouldn't the replacement maintain the same case?

```typescript
                /{{ (name) }}: {{ (convertedfieldtype)/i,
                "{{ $1 }}?: {{ $2");
```

bin/gql-gen.ts
```diff
@@ -40,53 +40,54 @@ async function main(): Promise<void> {
         const cwd = process.cwd();
         const lib = libDir(cwd);
 
+        // patch up codegen template to create optional properties
+        const codegenTemplate = path.join(cwd, "node_modules", "graphql-codegen-typescript-client", "dist", "index.js");
+        if (await fs.pathExists(codegenTemplate)) {
+            let contents = (await fs.readFile(codegenTemplate)).toString();
```
ddgenome: Should all this be in a try/catch block?

bin/gql-gen.ts
```diff
@@ -40,53 +40,54 @@ async function main(): Promise<void> {
...
+
         const graphqlFiles = await util.promisify(glob)(graphQlGlob);
+
         if (graphqlFiles && graphqlFiles.length > 0) {
-            gqlGenArgs.push(graphQlGlob);
+            config.documents = [graphQlGlob];
+            await generate(config);
         } else {
-            console.info("No GraphQL files found in project, generating default types");
+            console.info("No GraphQL files found in project. Skipping type generation...");
```
ddgenome: Are the default types brought in via a dependency?

bin/gql-gen.ts
cdupuis: Not sure. I’d like it to fail if this fails.

bin/gql-gen.ts
cdupuis: I’ll remove the i in the regex.

bin/gql-gen.ts
cdupuis: What default types? Without any graphql files, what types would you expect?

cdupuis:
@ddgenome any further thoughts on this?

cdupuis:
@ddgenome any more on this?

bin/gql-gen.ts
ddgenome: All the core types in the graph. Typically when there are no custom types in the project, you still get a types.ts with all core types Atomist defines. Will those be available some other way.

ddgenome:
Sorry, this slipped off my radar. I clarified my question.

bin/gql-gen.ts
cdupuis: No. They won’t.

cdupuis: You only get a generated types.ts when you actually define graphql queries or subscriptions. At that point you get a types.ts with the core types you used in your queries/subscriptions.

ddgenome: I think I am confusing custom schema and GraphQL files. You still create a types.ts if there is no custom schema, but you do not if there are no GraphQL queries/subscriptions/etc. Sorry for the confusion.

PR #483, opened 2019-03-06
WebSocket can be destroyed before message sent

ddgenome:
There were unhandled promise exceptions in the test because tests were
creating a websocket that never was connected and so was unable to
send messages.  The interval function kept running after the websocket
was destroyed, causing `this` to become undefined.  Since this could
also adversely affect normal operation, I added a check to ensure
`this` was defined.

Changes in the tests are whitespace and increasing the timeout on some
recently temperamental tests.

### DIFF
--- a/lib/internal/transport/websocket/WebSocketLifecycle.ts
+++ b/lib/internal/transport/websocket/WebSocketLifecycle.ts
@@ -1,5 +1,6 @@
 import * as TinyQueue from "tinyqueue";
 import * as WebSocket from "ws";
+import { logger } from "../../../util/logger";
 import { sendMessage } from "./WebSocketMessageClient";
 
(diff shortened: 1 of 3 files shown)


jessitron:
this is clearly better. Thank you. "this" as undefined... I don't understand quite how that happens at all and would like to (shouldn't it know that it's still referenced?). That can come later

ddgenome:
As I understand it, when the `setInterval` callback is created `this` is defined so `this.send` is dereferenced to the location of that method. Eventually, `this` goes out of scope and the object gets destroyed. Perhaps it wouldn't if the timer weren't `unref`ed. When the callback executes it is able to call the method at the `this.send` memory location but when that method gets entered, there is no `this` so the thing barfs.

PR #488, opened 2019-03-13
Adds branch to BitbucketRepoRef

### LINKED ISSUE #482: BitBucketServerRepoRef support cloning from specific branch
Add functionality to BitBucketServerRepoRef to support cloning (starting point) from a particular branch.   Currently, only master is supported.  Tested supplying this as SHA property but that doesn't not work currently.

lievendoclo:


ipcrm:
@lievendoclo I just tested this but I'm getting failures.  Will send you the log details.

PR #499, opened 2019-04-03
Add web identity and identity to source map

### LINKED ISSUE #497: Identity tokens in CommandHandlerRequest maps
Issue moved from atomist/sdm#702

Created by @slimslenderslacks at 2019-04-03T06:15:40Z:

The existing `source` map is augmented by an _optional_ `identity` map.  

```json
{
    "source": {
        "identity": {
            "token": "xxxxxxx"
        }
    }
}
```

The first workspace where you'll start to see identity maps in the source will be the atomist prod workspace using google auth.

### LINKED ISSUE #498: Support web user_agent in source of incoming command request
```
{
    "mapped_parameters": [],
    "secrets": [
        {
            "uri": "github://user_token?scopes=repo",
            "value": "8**************************************9"
        }
    ],
    "command": "ProvisionWorkspace",
    "source": {
        "user_agent": "web",
        "web": {
            "identity": {
                "sub": "cdupuis-test",
                "pid": "zjlmxjzwhurspem"
            }
        }
    },
    "correlation_id": "f4ad7b02-4956-4c90-8aaf-929c78218272",
    "team": {
        "id": "A6JJB10LQ",
        "name": "cd-test-01"
    },
    "automation": {
        "name": "@atomist/uhura"
    },
    "api_version": "1",
    "parameters": []
}
```

cdupuis:
and a couple of other smaller enhancements.

### DIFF
--- a/lib/internal/parameterPopulation.ts
+++ b/lib/internal/parameterPopulation.ts
@@ -17,7 +17,7 @@ import { Arg } from "./invoker/Payload";
  * @param args string args
  */
 export function populateParameters(instanceToPopulate: any, hm: CommandHandlerMetadata, args: Arg[]) {
-    args.forEach(arg => {
+    (args || []).forEach(arg => {
         if (arg.value !== undefined) {
             const parameter = hm.parameters.find(p => p.name === arg.name);
             if (parameter) {
--- a/lib/internal/transport/RequestProcessor.ts
+++ b/lib/internal/transport/RequestProcessor.ts
@@ -1,4 +1,3 @@
-
 import { HandlerResult } from "../../HandlerResult";
 
 /**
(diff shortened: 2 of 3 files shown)


lib/internal/parameterPopulation.ts
```diff
@@ -17,7 +17,7 @@ import { Arg } from "./invoker/Payload";
  * @param args string args
  */
 export function populateParameters(instanceToPopulate: any, hm: CommandHandlerMetadata, args: Arg[]) {
-    args.forEach(arg => {
+    (args || []).forEach(arg => {
```
lievendoclo: Good check, but that means someone explicitly defined that parameter as undefined? I would make this param optional then.

cdupuis: We are using the same pattern in that code already. That's why I went with this. Just staying consistent-

PR #513, opened 2019-04-12
Create single interface for creating RemoteRepoRef

ddgenome:
Each RemoteRemoteRef implementation has slightly different ways they
are created.  Add a single function with an object parameter to unify
creation of the various RemoteRepoRefs.

Export the new function and the missing BitBucketRepoRef and
ProviderType as ScmProviderType from the index.

Delint and fix token redacting and some tests.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018 Atomist, Inc.
+ * Copyright © 2019 Atomist, Inc.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


test/api/ApolloGraphClient.test.ts
```diff
@@ -16,7 +16,7 @@ import {
 
 describe("ApolloGraphClient", () => {
 
-    describe("api", () => {
+    describe.skip("api", () => {
```
jessitron: ?

ddgenome: Skip Apollo API tests
since GitHub token auth is no longer supported.

Issue #532, opened 2019-05-09
Send valid statsd metrics


ddgenome:
Currently the statsd listener implementation sends statsd metrics with tags, a dogstatsd extension, which are not accepted for standard statsd implementations. Adding tags should be optionals.

nebojsa-prodana:
I went forward and submitted a PoC in #533 to make statsd configuration more adaptable to user needs. Let me know whether this approach is aligned with your expectations.

PR #533, opened 2019-05-10
Expose integration point for different statsd client libraries

nebojsa-prodana:
Addresses #532.

Current state of things tie atomist statsd metrics to a specific statsd client library. Given that there is no standardization on some of the features used in here in statsd protocol like tagging and ecents, this means that users cannot consume metrics which are not compatible with dogstatsd datagram format.

The idea behind this PR is to democratize this and allow the users to specify a minimal required features adapter for statsd so that users can integrate their own statsd client library of choice.


package.json
```diff
@@ -64,7 +64,7 @@
     "app-root-path": "^2.1.0",
     "asciify": "^1.3.5",
     "async-exit-hook": "^2.0.1",
-    "axios": "0.19.0-beta.1",
+    "axios": "0.18.0",
```
nebojsa-prodana: I've changed this to `0.18.0` because `axios-mock-adapter` doesn't resolve with `0.19.0-beta-1` version specified making the test execution fail when run locally on my machine.

test/project/git/GitProject.test.ts
```diff
@@ -152,7 +152,7 @@ ding dong ding
         const gs2 = await p2.gitStatus();
         assert(gs1.sha === gs2.sha);
         await p.release();
-    }).timeout(5000);
+    }).timeout(10000);
```
nebojsa-prodana: I've had occasional test failures due to timeouts here, can revert if you'd like to keep the timeout as it was.

lib/util/statsdAdapter.ts
```diff
@@ -0,0 +1,64 @@
...
+    };
+}
+
+export interface StatsdAdapter {
+    increment(stat: StatsdAdapterStat, value: number, sampleRate?: number, tags?: StatsdAdapterTags, callback?: StatsdAdapterCallback): void;
+    timing(stat: StatsdAdapterStat, value: number, sampleRate?: number, tags?: StatsdAdapterTags, callback?: StatsdAdapterCallback): void;
+    gauge(stat: StatsdAdapterStat, value: number, sampleRate?: number, tags?: StatsdAdapterTags, callback?: StatsdAdapterCallback): void;
+
+    // TODO: This one is dogstatsd specific, as such it should be removed or made optional
+    event(stat: StatsdAdapterStat, text?: string, options?: {}, tags?: StatsdAdapterTags, callback?: StatsdAdapterCallback): void;
```
nebojsa-prodana: You are using dogstatsd events to track registrations, shutdowns and operational failures. This may require having a separate adapter going forward, to cleanly separate concerns and allow users to setup their own event ingression system.

package.json
ddgenome: I think we require the 0.19.0-beta.1 to fix various proxying issues. Although the failures you are seeing may be due to proxy issues.

lib/configuration.ts
```diff
@@ -283,6 +284,13 @@ export interface AutomationServerOptions extends AutomationOptions {
          * at the time of this writing.
          */
         port?: number;
+
+        /**
+         * adapterConfig. If not set, uses the default adapter which
+         * adapts to hot-shots statsd at the time of this writing.
+         */
+        adapterConfig?: StatsdAdapterConfig
```
cdupuis: I'd like to propose we rename this to `StatsClientFactory` and make it available in the configuration at `statsd.client.factory` to make it similar to other factories that create client implementations like http or ws. 

WDYT?

test/util/statsd.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            assert.doesNotThrow(() => {
+                const statsdListener = new StatsdAutomationEventListener(config);
+            });
+        });
+
+        it("works with node-statsd", () => {
+            const config = defaultConfiguration();
+            const expectedStatsdAdapterConfig = defaultStatsdAdapterConfig(config);
+
+            class NodeStatsDAdapter implements StatsdAdapter {
```
cdupuis: Can we make that a real class in the lib dir and make it the default? If people use the DataDog extensions they can configure the hot-shots based implementation.

lib/configuration.ts
nebojsa-prodana: Sounds good to me.

test/util/statsd.test.ts
nebojsa-prodana: That would be a breaking change in case that people are already using a datadog enabled statsd server, no? They might suddenly stop receiving the datadog events and not realize it.

cdupuis: That's true. Let's keep the current default then.

cdupuis:

* 3 approved reviews by @ddgenome, @cdupuis, @nebojsa-prodana
* 1 successful check

[atomist:generated] [auto-merge:on-check-success]

PR #535, opened 2019-05-13
Add toSubproject to InMemoryProject

jessitron:
this is driven by a need in the org-visualizer, as we add support for monorepos. I expect that we will find need for this elsewhere. For now I'm not worried about adding it to the test class, and we can think about the abstraction later


cdupuis:
Closing stale PR after a month.

Issue #543, opened 2019-05-21
Worker shutdown hooks do not complete


ddgenome:
Logs from a typical shutdown look like this:

```
2019-05-21T19:26:43.946Z [m:8] [info ] Shutdown initiated. Calling shutdown hooks
2019-05-21T19:26:43.948Z [w:15] [info ] Shutdown initiated. Calling shutdown hooks
2019-05-21T19:26:44.016Z [w:17] [info ] Shutdown initiated. Calling shutdown hooks
2019-05-21T19:26:44.016Z [m:8] [debug] Sending application event:
2019-05-21T19:26:44.016Z [w:17] [info ] Initiating worker shutdown
2019-05-21T19:26:43.949Z [w:15] [info ] Initiating worker shutdown
2019-05-21T19:26:44.065Z [m:8] [info ] Initiating WebSocket connection shutdown
2019-05-21T19:26:54.068Z [m:8] [info ] Closing WebSocket connection
2019-05-21T19:26:54.070Z [m:8] [info ] Shutdown hooks completed. Exiting...
```

These show a roughly ten second delay between the shutdown initiated and exiting. While the shutdown hook in the cluster master, shutting down the websocket, is initiated and shutdown, the shutdown of workers is initiated but never completed; the "Shutting down worker" log message never appears.

https://github.com/atomist/automation-client/blob/61c36874fdb7f75d832bdf9e29faa57dce31b158/lib/internal/transport/cluster/ClusterWorkerRequestProcessor.ts#L75

The timeout for the workers is 2500 ms longer than that for the websocket.

Perhaps when the master shuts down the workers are killed.

ddgenome:
When running locally on macOS, if I send SIGTERM to the main process, it appears the workers are sent a shutdown signal only after the main process completes its shutdown hooks:

```
2019-05-21T22:04:30.500Z [m:19162] [info ] Shutdown initiated. Calling shutdown hooks
2019-05-21T22:04:30.501Z [m:19162] [info ] Initiating WebSocket connection shutdown
2019-05-21T22:04:32.507Z [m:19162] [info ] Closing WebSocket connection
2019-05-21T22:04:32.508Z [m:19162] [info ] Shutdown hooks completed. Exiting...
2019-05-21T22:04:32.527Z [w:19163] [info ] Shutdown initiated. Calling shutdown hooks
2019-05-21T22:04:32.527Z [w:19164] [info ] Shutdown initiated. Calling shutdown hooks
```

[...]

PR #549, opened 2019-05-24
Rework cluster shutdown

ddgenome:
The master organizes shutdown of a cluster, waiting for the queue to
empty and telling the workers to shutdown.  If the workers receive a
shutdown signal, they check and potentially wait to see if the master
has sent them a shutdown message.  If not, they shutdown anyway but at
least give themselves some time to complete work.  Configure max
shutdown timeout based on configured grace period.  Execute all
shutdown hooks, even if some throw an error.

Provide helper functions for graceful termination.  Add graceful
functionality to non-cluster mode.

If not shutting down, replace any worker that exits, regardless of
how.

Provide a sleep function.

Delint files I touched.

Towards #543

Maybe `termination` should be a top-level configuration property.


cdupuis:

* 2 successful checks

PR #550, opened 2019-05-26
Improve cluster shutdown behavior

ddgenome:
Separate shutdown initiated behavior from replacing workers behavior
because shutdown can be initiated before you want to not replace
workers, e.g., when shutdown has been initiated but the work queue is
not empty.

Have the master send itself SIGTERM rather than call process.exit so
async shutdown hooks can execute properly.

Towards #543


cdupuis:

* 2 successful checks

PR #553, opened 2019-05-28
Allow for normal exit with async hooks

ddgenome:
Provide a means to process async shutdown hooks while returning a
specific exit status.  This is potentially breaking if user-provided
shutdown hooks using the default priority require that they execute
after the websocket is closed, which seems very unlikely.


cdupuis:

* 2 successful checks

PR #564, opened 2019-05-31
Allow configuration to be a `Promise<Configuration>`

cdupuis:
This makes creating `export const configuration` more flexible as it now come from an async method invocation.

### DIFF
--- a/lib/configuration.ts
+++ b/lib/configuration.ts
@@ -618,7 +618,7 @@ export function loadUserConfiguration(name?: string, version?: string): Automati
  * @param cfgPath location of automation configuration
  * @return automation configuration or undefined
  */
-export function loadAutomationConfig(configPath?: string): Configuration | undefined {
+export async function loadAutomationConfig(configPath?: string): Promise<Configuration | undefined> {
     let cfgPath = configPath;
     if (!cfgPath) {
         const cfgFile = "atomist.config.js";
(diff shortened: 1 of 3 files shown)


cdupuis:
Pull request rebase is in progress because @cdupuis pushed 3 commits to **master**:
- fa7a2e4 _Update axios dependency to tag latest_
- 8ba1901 _Fix compile error after TS upgrade_
- e586f4b _Auto merge pull request <https://github.com/atomist/automation-client/issues/565|#565> from atomist/automati..._

cdupuis:
Pull request rebase is in progress because @cdupuis pushed 1 commit to **master**:
- c054235 _Autofix: Third party licenses_

lib/configuration.ts
```diff
@@ -631,7 +631,10 @@ export function loadAutomationConfig(configPath?: string): Configuration | undef
     }
     if (cfgPath) {
         try {
-            const cfg = require(cfgPath).configuration;
+            let cfg = require(cfgPath).configuration;
+            if (cfg instanceof Promise) {
```
ddgenome: Is it possible for `require` to return a Promise?

ddgenome: Perhaps add a test that demonstrates it?

cdupuis: It is not the require that is the Promise. It is the configuration const that’s the Promise.

cdupuis: I’ll add a test good point. You can see this in action in the index of the new samples.

PR #597, opened 2019-09-05
Add support for code reloading without restart

cdupuis:
This adds a different `start-dev.js` script to the bin directory that starts the automation-client in a mode that reloads changes in the user code when there are changes to js files detected. 

As long as the user doesn't change metadata that would require a new registration to be made with the API, the changes are reloaded. If there are metadata changes (like changing the intent of a command handler) a diff is printed and the process exists.

This has no impact on running an automation-client with the current `start.js` script.

A corresponding PR on the atomist/cli will add a new `--dev` flag to `atomist start`.

Big thanks to @XertroV for the inspiration and some of the initial code.

### DIFF
--- a/bin/start-dev.ts
+++ b/bin/start-dev.ts
@@ -1,0 +1,106 @@
+#!/usr/bin/env node
+
+/*
+ * Copyright © 2018 Atomist, Inc.
+ *
+ * See LICENSE file.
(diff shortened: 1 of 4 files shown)


cdupuis:

* 2 successful checks

PR #602, opened 2019-09-19
Atm-gql-gen: accept optional 'input' and 'output' arguments

jessitron:
This retains existing functionality as a default, but lets people specify where their graphql directory is, and where they want the types to output.

This is helpful for moving the ts source files around to let clients debug into TS instead of .d.ts/js


cdupuis:
Turns out we don’t need this change as we found a different way to move TS files around.

PR #606, opened 2019-09-24
Introduce caching to Project API

johnsonr:
Makes project operations significantly faster.

- Caches file paths in `Project` to make successive glob and file searches much faster. This is available via the new `getFiles` method.
- Make `projectUtils` use the new method.
- Introduces user `cache` property in `Project` and `File` allowing expensive computations to be stored in a `Project` instance during its lifetime.
- Change `astUtils` to cache the AST for each parser by default.
- Add disk performance benchmark tests, skipped by default.

Also deprecates some `astUtils` functions, adding new functions that take an options object rather than individual parameters.

*Needs further testing before merging* (@lievendoclo), as well as review.

*DO NOT MERGE until further investigation. There may be a bug shown in aspect-sdm*

### DIFF
--- a/test/project/util/parseUtils.test.ts
+++ b/test/project/util/parseUtils.test.ts
@@ -185,7 +185,7 @@ describe("parseUtils", () => {
             .then(_ => {
                 // Check file persistence
                 const updatedFile = t.findFileSync(f.path);
-                assert(updatedFile.getContentSync() === initialContent.replace(oldPackage, oldPackage + "x"),
+                assert.strictEqual(updatedFile.getContentSync(), initialContent.replace(oldPackage, oldPackage + "x"),
                     `Content is [${updatedFile.getContentSync()}]`);
                 done();
             }).catch(done);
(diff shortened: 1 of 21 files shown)


test/tree/ast/typescript/typeScriptFileParserProject.test.ts
```diff
@@ -47,7 +47,7 @@ describe("TypeScriptFileParser real project parsing: TypeScript", () => {
                 assert(values.length > 2);
                 done();
             }).catch(done);
-    }).timeout(5000);
```
johnsonr: These changes don't relate to this PR, but were necessary due to slow internet while working on this.

test/project/util/parseUtils.test.ts
```diff
@@ -185,7 +185,7 @@ describe("parseUtils", () => {
             .then(_ => {
                 // Check file persistence
                 const updatedFile = t.findFileSync(f.path);
-                assert(updatedFile.getContentSync() === initialContent.replace(oldPackage, oldPackage + "x"),
+                assert.strictEqual(updatedFile.getContentSync(), initialContent.replace(oldPackage, oldPackage + "x"),
```
johnsonr: Style improvement

PR #614, opened 2019-10-03
Clean up tmp dirs on shutdown

### LINKED ISSUE #611: Register shutdown handler to remove temporary directories
Currently stable directories have a shutdown remover, but temporary directories do not.

ddgenome:
Rework temporary directory manager.  Move logic inside class.  Make
instance a singleton.  Defer initialization until first directory is
created.  Make sure interval reaper does not throw any errors, which
would result in an uncaught exception error.  Register shutdown
handler that deletes created temporary directories.  Add tests.

Close #611


cdupuis:

* 2 successful checks

PR #620, opened 2019-10-12
Consolidate on micromatch and fast-glob

ddgenome:
Replace uses of minimatch and multimatch with micromatch.  Use proper
options type for micromatch.  Replace glob, glob-stream, and globby
with fast-glob, the latter is just a wrapper around fast-glob that
provides features we do not use.  fast-glob uses micromatch.


cdupuis:
@ddgenome what do I have to do here?

ddgenome:
Nothing, the zenhub interface was not up to date and seeing the old data I made some changes that should not have been made.

PR #622, opened 2019-10-15
Invalidate file cache on content change

cdupuis:


lib/project/support/AbstractFile.ts
```diff
@@ -56,4 +57,8 @@ export abstract class AbstractFile extends AbstractScriptedFlushable<File> imple
 
     public abstract isBinary(): Promise<boolean>;
 
+    protected invalidateCache(): this {
+        _.keys(this.cache).forEach(k => delete this.cache[k]);
```
ddgenome:
```suggestion
        Object.keys(this.cache).forEach(k => delete this.cache[k]);
```

No need for lodash.

cdupuis:
There's a file name cache on the project level and a general content cache in the file. Those two things are completely different.

PR #624, opened 2019-10-18
Add exports to index

ddgenome:
Add isTokenCredentials and isGitHubRepoRef exports to index.


ddgenome:
The initial build failed but the retry worked. I'm not sure why the red X still shows for sdm/atomist/atomist-sdm.

Issue #225, opened 2018-02-23
Test errors with proper-lockfile 3.0.1


cdupuis:
https://travis-ci.org/atomist/automation-client-ts/builds/345171346

ddgenome:
Can't reproduce

cdupuis:
Upgrade to latest proper-lockfile.

ipcrm:
It appears this test failure was due to this breaking change in proper-lockfile: https://github.com/moxystudio/node-proper-lockfile/commit/b1d43e525418d74d6fa7cfa246e6451b3cbdd485

This was prior to the 3.0.1 release.

ipcrm:
After further review, it looks like the code that requires `proper-lockfile` can be deprecated.

PR #639, opened 2019-12-26
Remove CachingDirectoryManager (and proper-lockfile usage)

ipcrm:
Fixes https://github.com/atomist/automation-client/issues/225.


lib/spi/clone/CachingDirectoryManager.ts
```diff
@@ -110,26 +110,26 @@ type LockResult = LockAcquired | NoLockForYou;
 
 function pleaseLock(lockPath: string): Promise<LockResult> {
     return new Promise<LockResult>((resolve, reject) => {
```
ddgenome: It looks like the new version of lockfile.lock supports returning a promise, so wrapping in a promise seems unnecessary and this method could be greatly simplified.

test/project/git/LockfileLibrary.test.ts
```diff
@@ -7,14 +7,19 @@ import {
 
 describe("exploratory tests for proper-locking", () => {
```
ddgenome: This test seems to just test lockfile, not our code.

ipcrm: Indeed it does.  I don't believe it has any value - but I wasn't sure if it was put there to be a "canary in the mine" as it turned out to be in the failure from #225.  Should I just remove this?

lib/spi/clone/CachingDirectoryManager.ts
ipcrm: I'll take a crack at revising this, or removing, as mentioned https://github.com/atomist/automation-client/pull/639#pullrequestreview-336622281

ipcrm:
Updated to:
* Removed `pleaseLock` abstraction of `proper-lockfile`
* Remove unneeded tests
* Simplify logic using async/await
* Updated "clones a project subdirectory" test timeout to 15 seconds after having inconsistent executions that were taking 10-11 seconds and exceeding the existing timeout

lib/spi/clone/CachingDirectoryManager.ts
```diff
@@ -36,38 +37,37 @@ const cache = new StableDirectoryManager({
...
+                release: () => {
+                    logger.debug("Releasing lock on '%s'", existing.path);
+                    return lockResult().then(existing.release);
+                },
+                invalidate: () => {
+                    logger.debug("Invalidating '%s'", existing.path);
+                    return cache.invalidate(existing)
+                        .then(() => {
+                            logger.debug("Invalidated. Now releasing lock");
+                            return lockResult().then(existing.release);
```
ipcrm: I'm going to investigate `existing.release` further.  I"m not sure how this is actually consumed.

ipcrm:
@ddgenome @cdupuis 
Having dug into this further, I cannot find a place where this directory manager is even being used.  The repo loaders (https://github.com/atomist/automation-client/search?q=DefaultDirectoryManager&unscoped_q=DefaultDirectoryManager) are all making use of `DefaultDirectoryManager` which is defined here: https://github.com/atomist/automation-client/blob/3ea459ab3e1b171e1b3d02bd5ebb67c599284804/lib/project/git/GitCommandGitProject.ts#L53.  It's not clear to me how CachingDirectoryManager directory manager would get utilized unless the expectation is that it's only for custom repoLoaders.

In light of this, I'm proposing we simply remove it, along with the `proper-lockfile` package as this is the only place it's used.  That said, I'm not sure if there some usage I'm not aware of so looking for feedback.

ipcrm:
Having dug into this further, I cannot find a place where this directory manager is being used (outside tests).  The repo loaders (https://github.com/atomist/automation-client/search?q=DefaultDirectoryManager&unscoped_q=DefaultDirectoryManager) are all making use of `DefaultDirectoryManager` which is defined here: https://github.com/atomist/automation-client/blob/3ea459ab3e1b171e1b3d02bd5ebb67c599284804/lib/project/git/GitCommandGitProject.ts#L53  It's not clear to me how `CachingDirectoryManager` directory manager would get utilized unless the expectation is that it's only for custom repoLoaders.

In light of this, I'm proposing we simply remove it, along with the `proper-lockfile` package as this is the only place it's used.  That said, I'm not sure if there some usage I'm not aware of so looking for feedback.

package.json
```diff
@@ -132,6 +132,7 @@
   },
   "devDependencies": {
     "@types/mocha": "^5.2.7",
+    "@types/proper-lockfile": "^4.1.1",
```
ddgenome: This should be a regular dependency since automation-client is a dependency to other packages.

ddgenome:
It seems the behavior I am thinking about is provided by the CachingProjectLoader in sdm.

ipcrm:
Yea, as you mention with the CachingProjectLoader, it seems like the original intent of this directory manager ultimately got implemented elsewhere.

package.json
ipcrm: Updated.

cdupuis:
👍 for removing it.

